PR #3, opened 2014-03-06
ebuild: fix encoding of send_env

veelai:
Please review, I find encodings to be a pretty huge mess and thus am not sure,
that this is the correct solution. However it makes things work on my end.

_generate_env_str returned a unicode string in py3 which needs to be encodod
for serialization. This encodes the string if necessary. Note that ret prior
to encoding may either be bytes or unicode in py2 (it is unicode if data
contains a unicode string).

Rename _generate_env_str to reflect this change.

Note: In the first pull request I missed to delete a debug statement.

### DIFF
--- a/pkgcore/ebuild/processor.py
+++ b/pkgcore/ebuild/processor.py
@@ -620,7 +620,7 @@ class EbuildProcessor(object):
         else:
             self.write("set_sandbox_state 0")
 
-    def _generate_env_str(self, env_dict):
+    def _generate_env_data(self, env_dict):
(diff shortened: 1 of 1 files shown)


radhermit:
I'll probably do this along with the full py3 sweep I should be able to get to within the next month or so.

radhermit:
Sorry for the year+ delay but this should be fixed in edd18c1, I did it a bit differently leveraging some of snakeoil. Also it appears that pmerge works now in py3.

PR #4, opened 2014-03-11
ebuild-daemon: Exit on spurious commands during ebuild processing

veelai:
We are in an unknown state and thus should abort. One instance is the
command 'shutdown_daemon' which may arrive if there exists a bug causing
failure and thus a shutdown. If we do not exit the backtrace gets
unnecessarily clobbered.


radhermit:
Finally merged along with a few minor additions, sorry for the massive delay.

Issue #37, opened 2015-02-09
pmerge: document --quiet support in --help


austin987:

radhermit:
Closing this since the pullreq was merged. I'll look at adding more output squelching if the option is enabled later on or submit more issues with what you'd like quiet to output or not output.

austin987:
If I notice anything too verbose, I will. I didn't think it was supported at all at first glance, since I didn't see it in --help and so any other of portage's kitchen sink options weren't there either ;)

PR #38, opened 2015-02-09
pkgcore: add -q/--quiet options to help output

austin987:
Fixes https://github.com/pkgcore/pkgcore/issues/37


radhermit:
What do you want --quiet to actually do for pmerge? I assume just something similar to what portage does with emerge?

austin987:
Yeah, that's fine. It appears to be implemented, just not mentioned in --help.

radhermit:
Merged, I'll make it squelch more output later on.

PR #58, opened 2015-02-26
vdb: support missing /var/db/pkg dir

vapier:
When testing out an empty root, don't barf when the vdb doesn't exist.


radhermit:
Merged with munging to remove old virtuals support (and I seem to have forgotten to make sure you were still set as the commit author after rebasing, sorry).

PR #87, opened 2015-04-18
Fixed a 'pmerge --clean' behavior by setting replace=False.

flhs:
'clean' should reuse livefs repos.  With replace=True, livefs repos are
excluded by the _vdb_restrict in ebuild/resolver.py, causing the
resolver to generate a packets list that is not consistent with the
installed packets.

The error can be shown by running "pmerge --clean" then "pmerge -uD
@world".  Some packages (automake-1.15 in my case) removed by 'clean'
are reinstalled by 'update'.

Setting replace=False fixed the error.  We don't want to replace
anything by 'pmerge --clean' afterall.


radhermit:
Thanks! Merged.

PR #88, opened 2015-04-18
fixed function usev() in bash/eapi/common.lib

flhs:


ferringb:
Confirmed- nice catch.

PR #90, opened 2015-06-15
portage_conf: Handle git sync-uri= from repos.conf

mgorny:
Introduce a cheap hack to handle git sync-uri= in repos.conf. This isn't a replacement for full, proper and working repos.conf support. It's just enough to get 'pmaint sync' to work when git syncing is used.


radhermit:
I assume my recent repos.conf rework commit fixes this for you.

PR #91, opened 2015-06-23
Support building with setuptools when available

mgorny:
Import and use setuptools and its 'install' command when available. The setuptools monkey patching done by pip does not replace the install command, therefore the custom install collides with pip otherwise.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,8 +6,18 @@ import os
 import subprocess
 import sys
 
-from distutils import core, log, errors
-from distutils.command import build, build_scripts, install
+try:
+    import setuptools as core
+except ImportError:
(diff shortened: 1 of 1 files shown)


radhermit:
Did basically this in 605f79e with a few more changes.

Really I'd like to drop the entire snakeoil distutils extension usage so it's possible to do `pip install pkgcore` and get pip to install snakeoil as a dep. We'll leave that for another day though.

Thanks for this work.

mgorny:
Well, with the two PRs I sent I was able to do `pip install /home/mgorny/git/pkgcore` :). I'm going to try installing via git later.

PR #94, opened 2015-06-30
portage_conf: Do not prepend configroot to md5-cache path

mgorny:
Well, I'm not entirely sure about this one. But how it's done now, it means that it looks for `${repo}/metadata/md5-cache` and then uses `${root}/${repo}/metadata/md5-cache`, which means it doesn't use the actual cache…


radhermit:
Yeah this looks like some holdover from the past waiting for better prefix support as well.

radhermit:
Merged.

PR #95, opened 2015-06-30
ebd: set EPREFIX correctly

mgorny:
Set EPREFIX to just the correct prefix as required by the PMS rather
than some normalized combination of EROOT (which is supposed to contain
EPREFIX already...) and the prefix. This fixes non-prefix builds when
ROOT is non-null.


radhermit:
Thanks, I added a bit of simplification and fixes for few other edge cases for the final commit.

PR #98, opened 2015-07-02
repos: preserve supported profile formats when some are unsupported

zmedico:
For example, this makes it possible to work with a portage-2 format
profile which is mixed with an unsupported format such as
profile-bashrcs.


radhermit:
Merged in d69b123.

PR #99, opened 2015-07-09
Improvements for running as unprivileged user

mgorny:
Not the complete thing but collecting safe-to-merge changes here. Details in commit messages.


radhermit:
Merged. We probably should come up with a roadmap of some sort towards full prefix support.

PR #108, opened 2015-08-09
Cache fixes

mgorny:
These two make pkgcore-generated md5-cache useful for Portage. Well, it's not byte-exact still but it seems egencache is happy enough about it not to regenerate the whole cache.

### DIFF
--- a/pkgcore/cache/__init__.py
+++ b/pkgcore/cache/__init__.py
@@ -13,6 +13,7 @@ import os
 import operator
 
 from snakeoil import klass
+from snakeoil.chksum import get_handler
 from snakeoil.compatibility import raise_from
 from snakeoil.mappings import (
     ProtectedDict, autoconvert_py3k_methods_metaclass, make_SlottedDict_kls)
@@ -74,7 +75,7 @@ class base(object):
             return lambda data: '%.0f' % math.floor(data.mtime)
         # Skip the leading 0x...
         getter = operator.attrgetter(chf)
-        return lambda data: hex(getter(data))[2:].rstrip('L')
+        return lambda data: get_handler(chf).long2str(getter(data))
 
     @staticmethod
     def _get_chf_deserializer(chf):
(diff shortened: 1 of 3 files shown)


mgorny:
To be honest, I'm not very happy about the last commit but this is what egencache does. While Portage seems to be happy enough without it, egencache will still rewrite the cache file removing the value.

radhermit:
I'm not sure conforming with portage output is the best choice in many situations. For example, why would we want to sort the eclass list which would then lose the inherit order specified by the ebuild?

mgorny:
Well, I did it for readability. It was hard to compare a number of hashes when Portage and pkgcore sorted them differently. Not sure if inherit order is really relevant to the cache.

Conformity to a single standard would also extend the guarantee that various tools will use the cache entries rather than reject them. For example, if one did something as stupid as comparing the eclass list as strings rather than converting to dicts.

radhermit:
I think one of the problems is that the md5-dict cache format isn't actually specified formally anywhere, afaik, so this is another case of portage setting the "standard". If it is documented somewhere please point me to it.

mgorny:
Obviously it's not. Otherwise, I wouldn't be concerned so much about following Portage.

In any case, proper formatting for checksums is the obvious bug that caused cache rejection here. Eclass sorting seems not to be big issue though makes comparisons (i.e. validating pkgcore output) harder. The two last bits aren't necesary but egencache will rewrite cache files without them, and this will cause conflicts when people regenerate caches on top of repositories where the author uses different tool.

radhermit:
One question about the egencache implementation is why are they explicitly removing EAPI=0? See https://github.com/gentoo/portage/blob/master/bin/egencache#L300. Is that just for saving a few bytes? Because I can't come up with a better reason.

mgorny:
Yes, most likely. And I don't like that either. Maybe we could convince portage folks to remove that magic. Let's see...

radhermit:
Thanks, merged. I'm assuming you'll convince portage devs to remove their EAPI=0 magic in egencache. :)

PR #116, opened 2015-08-12
sync: Add git+svn syncer

mgorny:
Add a syncer for git+svn that uses the 'git svn' subcommand to mirror the complete history of a subversion repository.


radhermit:
Merged.

PR #117, opened 2015-08-13
pmaint regen: Support for use.local.desc & pkg_desc_index

mgorny:
To make pmaint a full replacement for egencache.

### DIFF
--- a/pkgcore/ebuild/repo_objs.py
+++ b/pkgcore/ebuild/repo_objs.py
@@ -122,8 +122,9 @@ class MetadataXml(object):
         self._longdescription = longdesc
         self._source = None
 
-        self._local_use = frozenset(
-            x.attrib['name'] for x in tree.findall('use/flag')
+        self._local_use = dict(
+            (x.attrib['name'], ' '.join(''.join(x.itertext()).split()))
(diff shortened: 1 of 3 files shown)


radhermit:
Merged with various changes including splitting out the support into separate funcs.

radhermit:
Ah, good call.

mgorny:
You broke return value handling :(.

        if options.use_local_desc:
            ret = update_use_local_desc(repo, out, err)
        if options.pkg_desc_index:
            ret = update_pkg_desc_index(repo, out, err)

Success in a latter call discards errors from former.

radhermit:
I realized that after I committed it but didn't feel like fixing it immediately.

radhermit:
Error handling should be cleaner now.

mgorny:
Still bad. Now pkg_desc_index update won't happen if use.local.desc had any issues.

radhermit:
I kind of prefer it that way if you're explicitly stating you want one or both, it should error out.

radhermit:
Never mind, I think I know what you mean. It's the metadata generation issue for bad ebuilds again I assume.

PR #118, opened 2015-08-17
sync/rsync: Clean up temporary timestamp file

mgorny:


mgorny:
@radhermit?

ferringb:
The intent here makes sense mind you- I'm just picking at the shitty logic/structuring I wrote there and hoping you find a cleaner flow for it.

mgorny:
@radhermit, ping. Can we merge this? I don't think anybody cares enough to improve it beyond that, tbh.

radhermit:
Timestamp handling via context manager done in 138e178528b901bdb74fb17c47fc3797c02d8833, left off on restructuring the logic because atm it makes no sense to me. :smile:

radhermit:
Made the logic a bit cleaner using a separate single file syncing class in 1b1885e2c80db99ce504e314c927f697341d6f2f.

PR #119, opened 2015-08-18
EbuildProcessor: move setpgid() calls to forked process

mgorny:
Call os.setpgid() in forked process (via new parameter to spawn()) to avoid race conditions between child being reaped and parent calling os.setpgid().

Seems to fix the issues I've reported.


radhermit:
As a sidenote, I've been thinking of moving most of the functionality of the spawn module to snakeoil.process.

mgorny:
Yeah, I've been positively surprised I need to patch only one package :P.

radhermit:
Merged along with some _exec() sig cleanup among other things.

PR #121, opened 2015-08-23
ebuild.digest: Set predictable mtime on Manifest files

mgorny:
Ensure that the Manifest's mtime is predictable by setting it to the
mtime of newest manifested file (either AUX, EBUILD, MISC or the old
Manifest itself). This allows us to cleanly preserve mtimes on
'unchanged' Manifests when regenerating them or converting from thin to
thick Manifests.

### DIFF
--- a/pkgcore/ebuild/digest.py
+++ b/pkgcore/ebuild/digest.py
@@ -22,6 +22,7 @@ from pkgcore.fs.livefs import iter_scan
 
 demandload(
     "errno",
+    "os",
     'snakeoil:mappings',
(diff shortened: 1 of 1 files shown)


mgorny:
I also wanted to switch to `AtomicWriteFile` but i'd rather wait till context manager support is fixed.

radhermit:
This has stalled somewhat since I wanted to fix AtomicWriteFile first, but haven't gotten around to that.

mgorny:
So far the code in Portage has caused so many issues, I think we shouldn't go this way.

PR #126, opened 2015-09-01
spawn.find_invoking_python: don't trust sys.executable

zmedico:
If the python interpreter is embedded in another process such as uwsgi,
then sys.executable is not reliable.

### DIFF
--- a/pkgcore/spawn.py
+++ b/pkgcore/spawn.py
@@ -492,16 +492,29 @@ def is_userpriv_capable(force=False):
     res = is_userpriv_capable.cached_result = (os.getuid() == 0)
     return res
 
+_invoking_python = None
+
 def find_invoking_python():
     # roughly... use sys.executable if possible, then major ver variations-
     # look for python2.5, python2, then just python, for example
+    # NOTE: sys.executable in unreliable if the interpreter is embedded
+    global _invoking_python
+    if _invoking_python is not None and os.path.exists(_invoking_python):
+        return _invoking_python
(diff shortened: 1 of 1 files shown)


zmedico:
In order to handle pythons that aren't cpython, I've updated the patch to test sys.executable with something like `[[ foo == $(python -c 'print("foo")' ]]`.

ferringb:
I'm pretty sure this can be simplified to a spawn_get_output call.  Barring
that, memozing this may be wise- as long as the access test succeeds in
subsequent invocations, then we can trust that the first test still holds
true.

Regarding the os.access call, Solaris is a bit weird for it:
http://bugs.python.org/issue14706 .  in years gone past I tried to protect
against that issue- that said, it may no longer matter.

zmedico:
I've updated it to use os.path.exists (instead of access) and spawn_get_output, and added memoization.

zmedico:
Updated to check if the memoized result still exists before returning it.

ferringb:
I'm pretty sure coverage will go down slightly from this, albeit wouldn't hugely matter in this case.  The change is fine by me- tim, care to do the honors since you own the codebase these days?

radhermit:
Honestly I'd like to make it so find_invoking_python isn't needed, i.e. the bash side doesn't call back into the python side via exec-ing pkgcore scripts but I guess this is fine to help fix the situation in the meantime.

radhermit:
Merged in c3e1df5.

PR #129, opened 2015-09-11
pmaint regen: Clean up stale cache entries

mgorny:
As the summary says. I have used the simplest solution possible, and it seems not to cause any significant slowdown.

One small issue is that cache entries for broken packages are not removed but that's more of a corner case to handle.

### DIFF
--- a/pkgcore/cache/flat_hash.py
+++ b/pkgcore/cache/flat_hash.py
@@ -124,13 +124,26 @@ class database(fs_template.FsBased):
         """generator for walking the dir struct"""
         dirs = [self.location]
         len_base = len(self.location)
+        # Note: the misc try/except clauses are to protect against concurrent
+        # modification of the cache resulting in transient errors.
         while dirs:
             d = dirs.pop(0)
+            try:
+                subdirs = os.listdir(d)
+            except EnvironmentError as e:
+                if e.errno != errno.ENOENT:
+                    raise KeyError(cpv, "failed accessing due to %s".format(e))
+                continue
             for l in os.listdir(d):
                 if l.endswith(".cpickle"):
                     continue
                 p = pjoin(d, l)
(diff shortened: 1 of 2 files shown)


mgorny:
There's also an issue that iterating over non-existing (empty) cache raises `OSError`. Should I change `__iter__` to handle non-existing directory as empty?

ferringb:
Yep.

mgorny:
Well, this abstraction is killing me. For 30 minutes, I'm trying to find the error handling code for cache updates. Or even figure out what is exactly happening when you expect the cache to be updated. So far I'm dozen levels deep in functions which basically don't do anything but call other functions, and I'm pretty sure I took a wrong turn.

So I have no clue how exactly does regen not raise exceptions. Neither do I agree with that behavior since we're pretty much wasting user's time doing regen which gets immediately discarded. But I'm going to just catch `ReadOnly` exception

ferringb:
https://github.com/ferringb/pkgcore/tree/clean-cache <-- this is a cleaned up version of what you're trying to do.

Regaring abstractions, parts of it are stupidly obtuse.  Other parts of it are intentionally forcing a separation so that the codebase doesn't wind up like portage did- everything reaching into each other, no clean/easy way to swap things out.  I'd suggest filing issues for the areas that are particularly nasty so that refactoring can be tracked/dealt with.

mgorny:
Thanks. Copied your code and fixed the typo. Is there anything else left to do?

ferringb:
@tim: stepping on on this one and merging it, feel free to yell if I'm taking liberties I shouldn't.

radhermit:
On 2015-09-15 23:08, ferringb wrote:
> @tim: stepping on on this one and merging it, feel free to yell if I'm taking liberties I shouldn't.

I'm generally fine with people doing work for me. ;)

PR #135, opened 2015-09-25
Bash 3.1 install

mgorny:
Long story short, I need to be able to install pkgcore on bash-3.1. This commit pretty much fixes the syntax errors that happen when trying to generate func list.


radhermit:
Merged.

PR #136, opened 2015-10-04
Fixed caching of some classes. Added caching to KeyedAndRestriction

neko259:


ferringb:
Looks sane offhand, although that's just a surface read...

neko259:
This does not fix all the issues. Much memory is eaten by choices class (or something similar, I don't remember much) that is not immutable because some "reduce" operation is performed on it sometimes. Code is far from perfect so I cannot fully understand how it works and how to reduce memory usage.

ferringb:
The memory explosion you're running into is described here https://github.com/pkgcore/pkgcheck/commit/3a5a8e2abe04bd543b305932f923072189eb159f

radhermit:
Thanks, merged. Sorry for the long wait.

PR #138, opened 2015-10-26
spawn: Ensure that SANDBOX_BINARY is actually Gentoo sandbox

mgorny:
Check the output of '${SANDBOX_BINARY} --version' to determine whether
sandbox is actually the Gentoo sandbox tool. For example, on Fedora
the path is used by SELinux sandbox which serves a different purpose
and causes hard-to-debug failures in pkgcore.


mgorny:
Note: tested on Gentoo only, waiting for Fedora results.

radhermit:
Merged with minor changes to make the check more similar to is_fakeroot_capable().

PR #142, opened 2015-11-06
portage_conf: Fix undefined var ref in profile error handling

mgorny:
Not sure if I'm using the most correct replacement, but it's the same as in `_find_profile_link()`.


radhermit:
Merged.

PR #147, opened 2015-11-10
Fix FD selection for bash process on Mac OS.

rhn:
File descriptors were derived from RLIMIT_NOFILE, yielding the value 255, which seems invalid on OSX. The solution uses consecutive numbers instead, since a platform-specific way to determine top FD numbers is unknown to me at the moment.

### DIFF
--- a/pkgcore/ebuild/processor.py
+++ b/pkgcore/ebuild/processor.py
@@ -349,15 +349,16 @@ class EbuildProcessor(object):
         spawn_opts["cwd"] = e_const.EAPI_BIN_PATH
         # little trick. we force the pipes to be high up fd wise so
         # nobody stupidly hits 'em.
+        # fds start from max-1 due to a bug in older bash
         max_fd = min(pkgcore.spawn.max_fd_limit, 1024)
         env.update({
-            "PKGCORE_EBD_READ_FD": str(max_fd-2),
-            "PKGCORE_EBD_WRITE_FD": str(max_fd-1)})
+            "PKGCORE_EBD_READ_FD": str(max_fd-3),
+            "PKGCORE_EBD_WRITE_FD": str(max_fd-2)})
         # pgid=0: Each ebuild processor is the process group leader for all its
         # spawned children so everything can be terminated easily if necessary.
         self.pid = spawn_func(
             [const.BASH_BINARY, self.ebd, "daemonize"],
-            fd_pipes={0: 0, 1: 1, 2: 2, max_fd-2: cread, max_fd-1: dwrite},
+            fd_pipes={0: 0, 1: 1, 2: 2, max_fd-3: cread, max_fd-2: dwrite},
             returnpid=True, env=env, pgid=0, *args, **spawn_opts)[0]
 
         os.close(cread)


pkgcore/ebuild/processor.py
```diff
@@ -347,17 +347,18 @@ def __init__(self, userpriv, sandbox, fakeroot, save_file):
         # force to a neutral dir so that sandbox/fakeroot won't explode if
         # ran from a nonexistent dir
         spawn_opts["cwd"] = e_const.EAPI_BIN_PATH
-        # little trick. we force the pipes to be high up fd wise so
-        # nobody stupidly hits 'em.
-        max_fd = min(pkgcore.spawn.max_fd_limit, 1024)
+
+        # Use consecutive FDs because deriving from RLIMIT_NOFILE yields unexpected results on OSX.
+        read_fd = 3
```
ferringb: This has a pretty strong chance of intersecting with ebuilds that screw with FDs, or build processes- they should behave better, but it happens (hence the original code).

how exactly is rlimit_nofile giving unexpected results?  That's a bit of a standard posix thing afaik- osx *should* handle it correctly.  I'm reasonably sure portage is still using this same trick- which I know worked on osx way back in the prefix days.

rhn:
I observed that rlimit_nofile=256, making the bash process get fds numbered 254 (pipe in), 255 (pipe out). It reads from 254 and processes the data, but on write to 255 there's an invalid FD error.
Manually setting the limit to 64 in pkgcore fixes this behaviour.
This leads me to thinking that rlimit_nofile doesn't tell the whole story and 255 is a special FD, making rlimit_nofile-1, 2, ... as wrong as X+1, 2, ...

pkgcore/ebuild/processor.py
mgorny: Just to be clear, I don't see any relevant code in Portage. The only occurence of NOFILE is used to scan for open fds, and this in turn seems to be used only to close them.

mgorny:
This is getting interesting. I did some experimentations, reproduced the issue… then it suddenly stopped failing… osx is ridiculous.

cbergstrom:
@OSX - lol

mgorny:
Well, I was wrong. I can now reproduce the pkgcore failure but so far my more and more complex test cases work fine. It's something that pkgcore must be doing to max_fd-1 somewhere… but so far, no clue what exactly.

mgorny:
Hallelujah, I found it! So it's bash-3.2 being at fault here. Early during startup, it opens the script for reading and then calls `move_to_high_fd (fd, 0, -1)` with the same goal as ours -- to avoid their fd colliding with random scripts. Sadly, the `0` here means that they don't check the fd and assume it's unused. AFAICS it has been fixed in newer bash versions and they started checking the fd. But since we support bash-3.2, we need to handle this.

Now, what bash exactly does in this case is:

1. it calls `getdtablesize()` which is equivalent to our rlimit call (at least on glibc & osx),
2. since `check==0`, it just takes the last fd (`getdtablesize()-1`).

Therefore, I guess that a good workaround would be to use `max_fd-3, max_fd-2` and leave `max_fd-1` for bash.

rhn:
Updated according to mgorny's instructions

radhermit:
Perhaps to make it somewhat less arbitrary we could start at the max limit and iterate downwards until we find two unused fds.

mgorny:
@radhermit that won't work. The point is, bash allocates `fd_max-1` after our script.

radhermit:
Ah, nevermind then.

ferringb:
Then shift it down by two, instead of one...

The point isn't to be the last in the space; the point is to squat in a
space that should strongly not collide.  Even if we /do/ collide, unless
it's bash itself (which we can't do much about), it just limits what can
talk to EBD.

PR #148, opened 2015-11-10
Stop forcing hardcoded system directories to be prepended to PATH.

rhn:
This allows the user to have control over which programs get found in PATH by setting own search order, in addition to some flexibility on systems with a nonstandard directory structure.

const.HOST_ROOT_PATHS isn't used anywhere else, therefore it's removed.


radhermit:
Merged with some comment changes in 0200ce74e771270c07a7e5c2568cb75ee0f60e5e.

PR #150, opened 2015-11-13
Compatibility fixes

mgorny:
Some compatibility fixes, for bash-3.2 and non-GNU toolset. More to come.


mgorny:
I need it for local deployment, plus there's probably some point in bootstrapping on systems where bash-3.2 is pre-installed. Crappy OSX, to be more precise. But let's close this one since the sed fix breaks all the fallbacks, and sed fails after all since the expression is too long for OSX sed, and I'll work on a better one.

PR #157, opened 2015-11-23
metadata: Include _eclass_exported_funcs mapping

mgorny:
This is my early work on making an explicit record of which phase function is used from which eclass, mostly for QA purposes. Since this is hard to obtain without rerunning ebuild, I'm storing the values in cache. Cache format and methods require review and better suggestions are welcome.

### DIFF
--- a/ebd/ebuild.lib
+++ b/ebd/ebuild.lib
@@ -518,12 +518,26 @@ __dump_metadata_keys() {
 	set +f
 
 	# defined phases... fun one.
-	local phases
+	local phases exp_funcs func tmp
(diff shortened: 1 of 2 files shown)


radhermit:
Was there some related bug for pkgcheck tests that should leverage this? Something like https://github.com/pkgcore/pkgcheck/issues/50?

mgorny:
Not really. It's mostly intended for diffing md5-cache before and after a change, e.g. to an eclass.

ferringb:
Still relevant?

PR #171, opened 2016-02-05
merge.triggers: Pass -X to ldconfig to avoid updating symlinks

mgorny:
Always use 'ldconfig -X' in order to prevent ldconfig from updating
SONAME symlinks. Those symlinks should already be correctly handled by
ebuilds, and letting ldconfig alter them may result in unintentional
'upgrade' of a symlink when newer library has not (yet) been cleaned up.

### DIFF
--- a/pkgcore/merge/triggers.py
+++ b/pkgcore/merge/triggers.py
@@ -286,7 +286,8 @@ class mtime_watcher(object):
 
 
 def update_elf_hints(root):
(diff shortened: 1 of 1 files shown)


mgorny:
Note: this is in line with what Portage does, and the change was done in Portage in order to correctly support downgrading with preserved-libs.

radhermit:
Thanks, merged as a single line, pkgcore stuff mostly accepts ~100 char limit lines.

PR #172, opened 2016-02-05
More OSX fixes

mgorny:
Still in testing phase. Fixes for missing OSX commands.

### DIFF
--- a/pkgcore/merge/triggers.py
+++ b/pkgcore/merge/triggers.py
@@ -36,6 +36,7 @@ demandload(
     'errno',
     'math:floor',
     'os',
+    "platform",
     're',
     'time',
     'snakeoil:process',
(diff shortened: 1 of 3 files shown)


pkgcore/merge/triggers.py
```diff
@@ -333,6 +334,10 @@ def _mk_ld_so_conf(self, fp):
             compatibility.raise_from(errors.BlockModification(self, e))
 
     def trigger(self, engine):
+        # ldconfig is only meaningful in GNU/Linux
+        if platform.system() != 'Linux':
```
mgorny: (this is what Portage does but it is incompatible with Jython which always returns `Java` x_x)

radhermit:
Trigger related commits merged, thanks.

I've reverted the SANDBOX_BINARY change I made back to the hardcoded path for now so it won't fail when it's missing on import. I think I'd rather thread that into the config framework as mentioned in #176 so all the binaries we search for can be handled similarly.

PR #178, opened 2016-02-14
test_commandline: Skip tty_detection_test on non-Linux as it hangs

mgorny:
Skip the tty_detection_test on non-Linux systems as the method used to
read pty causes the test to hang on non-Linux systems.


mgorny:
(reproduced both on Solaris and OSX)

PR #180, opened 2016-02-15
cpv.c: Replace glibc-specific rawmemchr() with simpler strlen() call

mgorny:
I see it as a bit cleaner alternative to #179.


radhermit:
So can #179 be closed then?

mgorny:
Yes, if you decide to apply this one.

radhermit:
Merged, thanks.

PR #181, opened 2016-02-21
Fix rmdir() to cover potential errno.EEXIST on non-empty dirs

mgorny:
Fix rmdir() uses to cover both the possibility of errno.ENOTEMPTY
and errno.EEXIST for non-empty directories. This behavior (either error)
is specified by POSIX, and Solaris returns EEXIST rather than ENOTEMPTY.

### DIFF
--- a/pkgcore/binpkg/repository.py
+++ b/pkgcore/binpkg/repository.py
@@ -350,7 +350,10 @@ class tree(prototype.tree):
         try:
             os.rmdir(pjoin(self.base, pkg.category))
         except OSError as oe:
-            if oe.errno != errno.ENOTEMPTY:
+            # POSIX specifies either ENOTEMPTY or EEXIST for non-empty dir
+            # in particular, Solaris uses EEXIST in that case
+            # https://github.com/pkgcore/pkgcore/pull/181
+            if oe.errno not in (errno.ENOTEMPTY, errno.EEXIST):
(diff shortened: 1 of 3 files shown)


mgorny:
Where's travis? I wanted to be lazy and avoid running tests locally ;-f.

ferringb:
You might want to add some comments to the code (or commit) explaining why EEXIST is warranted- I'm assuming you're doing it to address the scenario where the FS allows for hardlinks to a directory?

If that is the scenario you're trying to fix, it wouldn't surprise me if there are multiple other cases- might want to audit the code in whole for that.

mgorny:
No, this is just plain POSIX. From `man 3p rmdir`:

       [EEXIST] or [ENOTEMPTY]
                   The  path  argument names a directory that is not an empty directory,
                   or there are hard links to the directory other than dot or  a  single
                   entry in dot-dot.

mgorny:
I mean, there are no hardlinks there but POSIX says either code is fine.

ferringb:
Well, that's cracky.  Is it worth adding a comment in the code referencing the 'why' for this, since it's a nonobvious bit of POSIX?

mgorny:
Added a lot of comments everywhere ;-).

radhermit:
Thanks, merged.

Issue #122, opened 2015-08-23
pmaint digest: 'operations' object has no attribute 'build' on fetch failure


mgorny:
The following exception is thrown if `pmaint digest` is unable to fetch a distfile from any of listed SRC_URIs.

```
Error running an operation:
Traceback follows:
  File "/home/mgorny/git/pkgcore/pkgcore/util/commandline.py", line 919, in main
    exitstatus = main_func(options, out, err)
  File "/home/mgorny/git/pkgcore/pkgcore/scripts/pmaint.py", line 427, in digest_main
    if not repo.operations.digests(domain, options.query, observer=obs):
  File "/home/mgorny/git/pkgcore/pkgcore/operations/__init__.py", line 68, in _recast_exception_decorator
    return functor(*args, **kwds)
  File "/home/mgorny/git/pkgcore/pkgcore/operations/repo.py", line 262, in _cmd_api_digests
    domain, matches, observer, **options)
  File "/home/mgorny/git/pkgcore/pkgcore/ebuild/repository.py", line 79, in _cmd_implementation_digests
    if not pkg_ops.mirror(observer):
  File "/home/mgorny/git/pkgcore/pkgcore/operations/__init__.py", line 74, in _recast_exception_decorator
    compatibility.raise_from(exc_class(name))
  File "/home/mgorny/git/pkgcore/pkgcore/operations/__init__.py", line 68, in _recast_exception_decorator
    return functor(*args, **kwds)
  File "/home/mgorny/git/pkgcore/pkgcore/operations/format.py", line 130, in _cmd_api_mirror
    return self._mirror_op.fetch_all(self._get_observer(observer))
  File "/home/mgorny/git/pkgcore/pkgcore/operations/format.py", line 37, in fetch_all
    if not self.fetch_one(fetchable, observer):
  File "/home/mgorny/git/pkgcore/pkgcore/operations/format.py", line 51, in fetch_one
    self.failed_fetch(fetchable, observer)
  File "/home/mgorny/git/pkgcore/pkgcore/operations/format.py", line 58, in failed_fetch
    build_ops = self.domain.build_pkg(self.pkg, observer)
  File "/home/mgorny/git/pkgcore/pkgcore/config/domain.py", line 60, in build_pkg
    return self.pkg_operations(pkg, observer=observer).build(

OperationError:
 Unhandled exception in mirror
 'operations' object has no attribute 'build'

den4ix:
https://github.com/pkgcore/pkgcore/pull/194

PR #194, opened 2016-05-05
Improve exception handling in pmaint(pkgcore#113,#122)

den4ix:
Currently exceptions from fetch/base.py fetcher._verify are just swallowed in fetch/custom.py fetch and not handled properly. Also operations/format.py failed_fetch trying to call unexistent build property which causes error handling end up with things like following:
pmaint digest: 'operations' object has no attribute 'build' on fetch failure

### DIFF
--- a/ebd/generate_eapi_func_list.bash
+++ b/ebd/generate_eapi_func_list.bash
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 #
-# Generates a list of EAPI specific functions to avoid exporting to the saved
-# ebuild environment.
+# Generate the list of functions specific to an EAPI version.
 #
 # This script is run dynamically in a repo or tarball layout on initialization
 # of the build environment for each ebuild since different EAPIs require
--- a/ebd/generate_global_func_list.bash
+++ b/ebd/generate_global_func_list.bash
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 #
-# Generates a list of functions defined in the various bash support
-# libs to avoid exporting to the saved ebuild environment.
+# Generate the list of globally defined functions.
 #
 # This script is run dynamically in a repo or tarball layout on initialization
 # of the build environment. For installed versions, a static function list is


pkgcore/operations/format.py
```diff
@@ -45,10 +45,11 @@ def fetch_one(self, fetchable, observer):
         # XXX hack atm, could use better logic but works for now
         try:
             fp = self.fetcher(fetchable)
-        except Exception:
+        except Exception as e:
+            observer.error(str(e))
+            observer.error("failed fetching %s", fetchable)
             fp = None
         if fp is None:
-            self.failed_fetch(fetchable, observer)
```
radhermit: This line can't be dropped, otherwise the pkg_nofetch phase doesn't get run properly on failures. See man 5 ebuild or PMS (app-doc/pms) for pkg_nofetch info.

den4ix: Seems like nofetch just wasn't executed before =)

radhermit: I made it work ~3 years ago with 76747f. If you're talking about forgetting to only run it for fetch restricted packages as per the spec, then yes I just noticed and fixed that. :smile:

den4ix: Yeah, that's what I meant :). Seems like there only two cases, one is:
NEWS.rst:162:- Run pkg_nofetch phase when any files in SRC_URI fail to be fetched.
and another one if you want to run pebuild <atom> nofetch,  right ?

radhermit:
Can you rebase this on current master?

Also, it seems like this will check uris across all available third party mirrors instead of failing out after X attempts (default currently being 10). I don't think that's what we want as some third party mirrors have way too many mirrors with many of them being slow to respond from almost anywhere in the world.

pkgcore/operations/format.py
```diff
@@ -55,7 +56,11 @@ def fetch_one(self, fetchable, observer):
         return True
 
     def failed_fetch(self, fetchable, observer):
-        build_ops = self.domain.build_pkg(self.pkg, observer)
```
den4ix: I'm not sure this line will work. As it passes raw pkg to build_pkg, it will not generate needed build operation from src_operation(build_operations). Seems like it needs PackageWrapper, so that it can call binded _generate_pkg_operations from _ConfiguredTree instead, which gets buildable and run actual nofetch on it.

pkgcore/operations/format.py
```diff
@@ -57,7 +58,9 @@ def fetch_one(self, fetchable, observer):
     def failed_fetch(self, fetchable, observer):
         # run pkg_nofetch phase for fetch restricted pkgs
         if 'fetch' in self.pkg.restrict:
-            build_ops = self.domain.build_pkg(self.pkg, observer)
```
den4ix: I'm not sure this line will work. As it passes raw pkg to build_pkg, it will not generate needed build operation from src_operation(build_operations). Seems like it needs PackageWrapper, so that it can call binded _generate_pkg_operations from _ConfiguredTree instead, which gets buildable and run actual nofetch on it.

den4ix: I'm not sure this line will work. As it passes raw pkg to build_pkg, it will not generate needed build operation from src_operation(build_operations) and end up with same error in domain:build_pkg, like no build operation . Seems like it needs PackageWrapper, so that it can call binded _generate_pkg_operations from _ConfiguredTree instead, which gets buildable and run actual nofetch on it.

radhermit: Right, I meant rebase your fix for that on top of my fix for only running pkg_nofetch when an ebuild is fetch restricted. I was never trying to fix this bug with my previous commit.

den4ix: Oh, a little misunderstanding =) Anyway, I guess it's ready for merge and you might close those two bugs.

PR #203, opened 2016-05-20
fix error output for bad tarballs or manifests

### LINKED ISSUE #113: pmaint digest: fix error output for bad tarballs or manifests
If one of the chksums for a tarball in the manifest doesn't match things currently appear to explode. We should raise something like ChecksumFailed and output a nice error message in pmaint digest.

4eetah:
Fixes #113


arthurzam:
Those changes were done differently in the repo, so nothing to import from this PR now.
Thank you for your work.

Issue #213, opened 2016-07-01
pkgcore no longer complains about no repo_name


mgorny:
It seems that pkgcore has stopped complaining when repository doesn't have profiles/repo_name. Looking via the API, `repo_id` is set to match config section, and `repo_name` is always empty (even on repos with repo_name).

mgorny:
Oh, I see 0d6311b52b06bfc1d31db40486ada443e7230c6d being at fault… I wonder how many new breakages I will discover after reverting it.

mgorny:
Or maybe I should just add another attribute for the contents of that file. Probably pkgcore relies on repo_id being set internally.

radhermit:
Re-added the warning, but note that two warnings for each unnamed repo will still be thrown. I have to figure out what's causing configs to be parsed twice and see if that can be fixed easily.

Also, attrs already existed and still do for the separate names. Use repo_name for profiles/repo_name or the repo-name field in metadata/layout.conf that overrides it, and config_name for the user specified repo name in repos.conf.

Finally repo_id defaults to config_name as that makes it more consistent for a user to override repo names without having to change the relevant files/settings in the repo itself.

PR #214, opened 2016-07-01
RepoConfig: add an explicit pms_repo_name attr for profiles/repo_name

mgorny:


radhermit:
I'm pretty sure this should be fixed via the commit related to #213. Use the repo_name attr to respect what the repo calls itself and use repo_id for whatever the user has possibly overridden it with in repos.conf settings.

mgorny:
Except that once again you've completely missed the point. I need to be able to get *both* the name from `profiles/repo_name` and from `metadata/layout.conf`, in order to verify that they match and throw an explicit QA error if they do not.

radhermit:
I guess I don't see that being a big QA issue and imo there should be one place where repo names are set in the repo, not two, but we're already well past config sanity ;).

PR #217, opened 2016-07-28
replace conditional restrictions accumulation in transitive_use_atom

4eetah:
partly fixes pkgcore#80, futher thougths on separation of
conditional restrictions may be required

### DIFF
--- a/pkgcore/ebuild/atom.py
+++ b/pkgcore/ebuild/atom.py
@@ -11,6 +11,7 @@ gentoo ebuild atom, should be generalized into an agnostic base
 __all__ = ("atom", "transitive_use_atom")
 
 import string
+from itertools import product
 
 from snakeoil.compatibility import is_py3k, cmp, raise_from
 from snakeoil.demandload import demandload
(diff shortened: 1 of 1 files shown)


den4ix:
not completed yet, it turned out that some depset code still using old approach, so it needs a bit more efforts

pkgcore/ebuild/atom.py
```diff
@@ -11,6 +11,7 @@
 __all__ = ("atom", "transitive_use_atom")
 
 import string
+from itertools import product
```
ferringb: Background context, I originally wanted to use itertools.product, pkgcore just didn't yet support those versions of python.

That said, unless I missed something, converting to itertools here just eliminates the OOM; however, this codepath should still effectively hang any consuming code due to the exponential explosion.

Take a look at https://github.com/pkgcore/pkgcheck/commit/3a5a8e2abe04bd543b305932f923072189eb159f for background context; that atom (which isn't uncommon) is 2^25 possible permutations.

Your usage of product *will* suppress the OOM, but it would allow any code that isn't protected to go down that expoentential pathway, generating each possible permutation.  For pmerge, the resolver likely will cache those results anyways, resulting in OOM.

That said, assuming this code passes tests, your version is a helluva lot fucking cleaner than mine, so merging on that alone is probably wise; I just want to be clear that it shifts the problem from one axis to another (or at least, that's what I'm reading in the code while doing a 5m drive by review).

den4ix: Thanks =) Doesn't pmerge always redirect all the work to the transitive_use_atom.evaluate_conditionals ? which then just calculate conditional use dependency flags based on use flags provided by a parent. Or have I missed something I suspect might be related to 2 ?
I only see usage of that functionality in two places:
1. pquery while printing the revdeps of the package and intersecting with generated atoms(with all possible permutations of conditional use flags)
2. requesting for attribute change in package wrapper and then attempt force_{True,False} essentially trying to make corresponding restriction True or False.
Could you provide some examples where the second case might be applied/needed ? like forcing some attribute to change it's state despite having some initial configuration.

ferringb: On Fri, Jul 29, 2016 at 4:08 PM, Denis Romanchuk <notifications@github.com>
wrote:

> In pkgcore/ebuild/atom.py
> <https://github.com/pkgcore/pkgcore/pull/217#discussion_r72873036>:
>
> > @@ -11,6 +11,7 @@
> >  __all__ = ("atom", "transitive_use_atom")
> >
> >  import string
> > +from itertools import product
>
> Thanks =) Doesn't pmerge always redirect all the work to the
> transitive_use_atom.evaluate_conditionals ? which then just calculate
> conditional use dependency flags based on use flags provided by a parent.
> Or have I missed something I suspect might be related to 2 ?
> I only see usage of that functionality in two places:
> 1. pquery while printing the revdeps of the package and intersecting with
> generated atoms(with all possible permutations of conditional use flags)
> 2. requesting for attribute change in package wrapper and then attempt
> force_{True,False} essentially trying to make corresponding restriction
> True or False.
> Could you provide some examples where the second case might be
> applied/needed ? like forcing some attribute to change it's state despite
> having some initial configuration.
>

I don't have anything on hand, and these days, anything I comment, it's
best to assume I'm working from past knowledge.  I'm heavily fallible, in
other words. :)

What you're describing does sound about right- that it generated the
intermediate expanded conditional form, then evaluated it to calculate down
to a locked form of the dependencies.  Assuming that's the case, then the
only issue would be the 2^25 item walk- OOM shouldn't be possible there.

I'd still double check that I didn't add some caching in for this, but I
think you're likely right.

devurandom:
Did this progress any further since July/August? I am hitting the OOM right now and am thus interested in a solution.

den4ix:
Thanks for being interested, perhaps I should renew some work on pkgcore and eventually finish this PR even though it's a bit tricky one, any help is welcome, of course =)

ferringb:
@arthurzam my suspicion is this PR is no longer relevant due to the changes you/others made to the transitive use parsing.  If I'm wrong on that, please reopen this PR for revisiting.

Issue #154, opened 2015-11-23
pmerge: handle permission failures in a more user friendly way


radhermit:
We should probably check for permissions soon after launch as well as fail out in a more friendly fashion.

PR #220, opened 2016-07-31
pmerge: handle permission failures in a more friendly way

den4ix:
Issue #154

### DIFF
--- a/pkgcore/merge/triggers.py
+++ b/pkgcore/merge/triggers.py
@@ -73,7 +73,7 @@ class base(object):
     _engine_types = None
     priority = 50
 
-    suppress_exceptions = True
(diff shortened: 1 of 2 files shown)


den4ix:
I don't really know if you would like "suppress_exceptions = False" for all triggers. Another way to propagate such exceptions is checking for errno.EACCES in merge engine and having suppress_exceptions enabled:
```python
345                     except Exception as e:
346                         if not trigger.suppress_exceptions:
347                             raise
348
349                         handle = stringio.text_writable()
350                         traceback.print_exc(file=handle)
351
352                         self.observer.warn(
353                             "unhandled exception caught and suppressed:\n%s", handle.getvalue())
```

PR #223, opened 2016-10-27
[glsa] Fix handling ranges in GLSAs

mgorny:
Fix the following issues in GLSA range handling:

- rgt with -r0 is perfectly valid (it matches -r1+),
- rle with -r0 is =, not ~,
- the glob code does not handle le, lt, gt, ge, so reject those.


mgorny:
Note that I haven't really tested it; but I can confirm that this fixes the spurious errors with rgt ...-r0 ranges.

radhermit:
I don't suppose you'd like to add some tests for the changes? :smile:

mgorny:
I guess that shouldn't be hard.

mgorny:
And here's version with extra unit tests. I've decided not to assert for exception in 'rlt -r0', and instead 'convert' it implicitly into an always-false restriction.

PR #230, opened 2017-02-01
ebuild/profiles: support profile-formats = profile-set

zmedico:
Add minimal syntax support for profile-formats = profile-set in profile
packages files. In the portage implementation, @profile packages do
not limit parallelization of builds in the way that @system packages do.

@chutz


zmedico:
Test coverage is fixed now.

PR #232, opened 2017-02-26
Disable sandbox when pkgcore is run inside sandbox

mgorny:
Sandbox prohibits starting another sandbox inside a running one, which
in turns causes pkgcore to error out hard. To avoid this, just disable
sandbox when another sandbox is detected. The rule for that is copied
from sandbox itself.


radhermit:
Merged, thanks.

PR #233, opened 2017-03-03
ebuild/profiles: support -* wildcard in system set

zmedico:
Support -* in order to make it easier to create profiles for minimal
systems (especially those built entirely from binary packages). This
form of wildcard is supported by portage 2.3.4 (commit 
https://github.com/gentoo/portage/commit/f4b0714010f1237280fd48dcb65989679917d20a for [bug 610670](https://bugs.gentoo.org/show_bug.cgi?id=610670)).

### DIFF
--- a/pkgcore/ebuild/profiles.py
+++ b/pkgcore/ebuild/profiles.py
@@ -157,9 +157,12 @@ class ProfileNode(object):
         profile_set = repo_config is not None and 'profile-set' in repo_config.profile_formats
         # sys packages and visibility
         sys, neg_sys, vis, neg_vis = [], [], [], []
+        neg_sys_wildcard = False
         for line in data:
             if line[0] == '-':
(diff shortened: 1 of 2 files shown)


radhermit:
Is this support explicitly noted in any documentation? All I can find is the note under the related packages section in the portage man page that mentions prefixing atoms with '-'. If globs are supported it would be nice to note that somewhere there as well.

zmedico:
I've opened https://github.com/gentoo/portage/pull/166 to document it in the portage.5 man page. Looks good?

radhermit:
Yeah, that looks fine. Thanks. :smile:

radhermit:
Merged in 2254a440afa966cbb565c00bbf475811cb1e75e8 along with a couple extra tests and parameter to explicitly enforce collapsed clearing to only the system set.

PR #235, opened 2017-07-16
ebuild.conditionals: Do not collapse nested AndRestrictions

mgorny:
Disable collapsing nested AndRestrictions and preserve the exact structure of the input. While technically the nested and collapsed variants are equivalent, collapsing them makes it impossible to detect them in pkgcheck and report as a meaningless construct.

*Note*: I'm not using pkgcore as my packages manager. I'm testing it on pkgcheck but I'd appreciate if someone could confirm it doesn't hit some unlikely corner case in the PM part.


radhermit:
Thanks, merged.

PR #236, opened 2017-09-03
test_conditions: Fixed parameter order in output

mgorny:


radhermit:
Merged in b4b27c13425d2665e1fe16726b0e436551cb14fe.

PR #239, opened 2017-09-09
portage_conf: Ignore backup files (*~) in Portage configs

mgorny:
Ignore backup files (that is, files ending with a tilde) while
processing Portage configuration directories. Some of the configuration
manipulation tools (flaggie, eselect-repository) rely on this behavior
to create backups, and pkgcore currently reads them as additional
(obsolete) configuration.

### DIFF
--- a/pkgcore/ebuild/portage_conf.py
+++ b/pkgcore/ebuild/portage_conf.py
@@ -269,6 +269,8 @@ def load_make_conf(vars_dict, path, allow_sourcing=False, required=True,
         sourcing_command = 'source'
 
     for fp in sorted_scan(os.path.realpath(path), follow_symlinks=True, nonexistent=True):
+        if fp.endswith('~'):
+            continue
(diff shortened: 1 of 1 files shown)


radhermit:
Doesn't this only make.conf and repos.conf parsing? Don't you want the same thing to be allowed for everything under /etc/portage?

mgorny:
It's your project, you should know xD. But yes, we'd want it for all files and I think it was supposed to solve it but maybe I was wrong. Don't recall anymore.

radhermit:
It was a rhetorical question as I was confused why you left that part out. :smile:

radhermit:
Should be fixed in 56a1e224c8859bb1f9f1ffa0a56238eaf82c46be slightly differently by extending the underlying scan funcs. Note that this will also ignore hidden/backup files for repo profile files as well too for the time being.

PR #240, opened 2017-09-09
Make external commands fatal during metadata regen

mgorny:
Per the PMS, calling external commands during metadata regen is forbidden. Use `command_not_found_handle()` combined with forcing invalid `PATH` to catch them and die on them. I've also removed QA interceptors since all the listed commands are caught implicitly now.

All the failures in Gentoo were fixed already, and other repositories are WIP. It's also running on CI for a few days now.


radhermit:
Nice work, merged.

It could be interesting to go a step further and use seccomp filtering to lock things down even more so using fully path-ed executables would fail out.

PR #246, opened 2018-04-27
Minimal EAPI 7 support (for cache gen)

mgorny:
Here's minimal implementation of EAPI 7 (as noted in #247). It passes tests for me and seems to work correctly with a dummy local ebuild (with and without BDEPEND).


mgorny:
FTR, it seems that EAPI 7 package.mask as a directory causes pkgcore to crash hard.

radhermit:
Merged, remaining EAPI 7 support for regular pm usage will slowly filter in over time.

PR #248, opened 2018-05-05
Fix listing pkg_config & pkg_nofetch in DEFINED_PHASES

mgorny:
Fix the code setting DEFINED_PHASES.  Change pkg_configure
to pkg_config, and add missing pkg_nofetch.


radhermit:
Merged, thanks.

PR #258, opened 2019-03-09
edb/eapi/common.bash: Add "test" to pre-phase check for WORKDIR fallback

ulm:
This was omitted from the original EAPI 4, but retroactively fixed
by the following PMS commit:
https://gitweb.gentoo.org/proj/pms.git/commit/?id=0038f90a942f0856ae2533b26f709002a3ec80ae


radhermit:
Merged in https://github.com/pkgcore/pkgcore/commit/10cdad6cacb5fca562309b217f36cd459ef6d8c7

PR #263, opened 2019-07-26
Support PROPERTIES=live as live ebuild indicator

mgorny:


mgorny:
CC @ulm

src/pkgcore/ebuild/ebuild_src.py
```diff
@@ -314,7 +314,8 @@ def mandatory_phases(self):
 
     @property
     def live(self):
-        return bool(const.VCS_ECLASSES.intersection(self.inherited))
+        return ('live' in self.properties or
+                bool(const.VCS_ECLASSES.intersection(self.inherited)))
```
ulm: Why keep the eclasses here?

radhermit: In part because pkgcheck was using it as well for tests.

radhermit:
Thanks, merged with some changes.

PR #264, opened 2019-08-02
repo_objs: Fix missing f-string for maintainer with desc

mgorny:


radhermit:
Thanks, merged with some tests to check maintainer description rendering.

PR #265, opened 2019-08-02
Add support for processing projects.xml

mgorny:


src/pkgcore/ebuild/repo_objs.py
```diff
@@ -170,6 +171,216 @@ def __init__(self, metadata_xml, manifest):
...
+
+    def __init__(self, email, name=None, role=None, is_lead=None):
+        if email is None:
+            raise ValueError('email for project member must not be null')
+        self.email = email
+        self.name = name
+        self.role = role
+        self.is_lead = is_lead
+
+    def __eq__(self, other):
```
mgorny: FTR, I've tried using `klass.generic_compare` but it didn't work. Does it work anymore with py3?

mgorny: `s/generic_compare/generic_equality/`

src/pkgcore/ebuild/repository.py
```diff
@@ -400,6 +400,10 @@ def __init__(self, location, eclass_cache=None, masters=(), cache=(),
             self, cache, self.eclass_cache, self.mirrors, self.default_mirrors)
         self._shared_pkg_cache = WeakValCache()
         self._masked = RestrictionRepo(repo_id='masked')
+        self.projects_xml = repo_objs.LocalProjectsXml(
+            pjoin(self.location, 'metadata', "projects.xml"))
+        import IPython
```
radhermit: Errant debugging lines?

tests/module/ebuild/test_repo_objs.py
```diff
@@ -109,6 +109,198 @@ def test_longdesc(self):
...
+      <email>dev2@example.com</email>
+    </member>
+    <member>
+      <email>dev5@example.com</email>
+    </member>
+    <subproject ref="recursive-subprojects@example.com" inherit-members="1"/>
+  </project>
+</projects>"""
+        return repo_objs.ProjectsXml(data_source(s.encode('utf-8')))
+
```
radhermit: Probably need a basic test for the case where projects.xml doesn't exist or is empty. I'd make get_projects_xml() accept a string param defaulting to None in which case it gets the string already defined there. Then have a test pass it the empty string and check viability, etc.

src/pkgcore/ebuild/repo_objs.py
radhermit: I think it still works, how were you using it or what issues were you running into? It needs to be set as a metaclass and then you need to define __attr_comparison__ properly.

src/pkgcore/ebuild/repository.py
mgorny: No, entirely expected behavior ;-).

tests/module/ebuild/test_repo_objs.py
mgorny: I don't think we should accept empty `projects.xml`. Non-existing sure but empty is invalid XML, isn't it?

src/pkgcore/ebuild/repo_objs.py
mgorny:
```
    __metaclass__ = klass.generic_equality
    __attr_comparison__ = ('email', 'name', 'role', 'is_lead')
```

and the tests fail comparison, while they work with explicit `__eq__`.

tests/module/ebuild/test_repo_objs.py
radhermit: Sure, but I'd still make sure the parsing mechanism outputs an error/warning for that case and ignores it instead of throwing an exception and outputting a backtrace. I've been trying to make pkgcore more acceptable of bad input and logging/ignoring it so then pkgcheck can catch/flag those cases.

src/pkgcore/ebuild/repo_objs.py
radhermit: py3 changed how you specify metaclasses in [PEP 3115](https://www.python.org/dev/peps/pep-3115/). What happens if you switch to that format?

tests/module/ebuild/test_repo_objs.py
mgorny: Ok, made it lenient with `logger.error()`.

src/pkgcore/ebuild/repo_objs.py
mgorny: Yep, that works.

mgorny:
Applied all suggestions.

src/pkgcore/ebuild/repo_objs.py
```diff
@@ -170,6 +171,217 @@ def __init__(self, metadata_xml, manifest):
...
+    :type is_lead: C{bool}
+    :ivar is_lead: whether the member is a project lead.
+    """
+
+    __slots__ = ('email', 'name', 'role', 'is_lead')
+    __attr_comparison__ = ('email', 'name', 'role', 'is_lead')
+
+    def __init__(self, email, name=None, role=None, is_lead=None):
+        if email is None:
+            raise ValueError('email for project member must not be null')
```
radhermit: Raising ValueError for each object during instantiation is fine, but I'd prefer then that it be caught, logged as an error, and the related entry skipped when parsing projects.xml if possible. Otherwise, an invalid projects.xml file will cause premature exits when a user possibly doesn't even care about projects.xml, e.g. pkgcheck will error out and throw a backtrace when running the related check when instead it should probably throw an invalid projects.xml warning and continue running the rest of the checks.

Along those lines, it could help to add some basic tests for invalid projects.xml to codify what should occur on certain types of bad input.

src/pkgcore/ebuild/repo_objs.py
```diff
@@ -170,6 +171,217 @@ def __init__(self, metadata_xml, manifest):
...
+
+    def __init__(self, ref, projects_xml, inherit_members=None):
+        if ref is None:
+            raise ValueError('ref for subproject must not be null')
+        self._ref = ref
+        self.inherit_members = inherit_members
+        self._projects_xml = projects_xml
+        self._cached_project = None
+
+    @property
```
radhermit: It seems you could use @klass.jit_attr here as that's the behavior you're approximating.

src/pkgcore/ebuild/repo_objs.py
```diff
@@ -170,6 +171,217 @@ def __init__(self, metadata_xml, manifest):
...
+    def _project(self):
+        if self._cached_project is None:
+            self._cached_project = self._projects_xml.projects[self._ref]
+        return self._cached_project
+
+    __getattr__ = klass.GetAttrProxy("_project")
+    # TODO: radhermit, why doesn't this work? because of __slots__?
```
radhermit: Possibly because of __slots__, yes. Make _project a jitted attr, make sure to add it's underlying attr name to __slots__ (jitted attrs by default cache their values at '_' + func_name) and it should work.

src/pkgcore/ebuild/repo_objs.py
```diff
@@ -170,6 +171,217 @@ def __init__(self, metadata_xml, manifest):
...
+                kwargs[k] = p.findtext(k)
+
+            members = []
+            for m in p.findall('member'):
+                m_kwargs = {}
+                for k in ('email', 'name', 'role'):
+                    m_kwargs[k] = m.findtext(k)
+                m_kwargs['is_lead'] = m.get('is-lead', '') == '1'
+                members.append(ProjectMember(**m_kwargs))
+            kwargs['members'] = tuple(members)
```
radhermit: I'd move the tuple() to the related object instantiation so the related object attr is always stored as an attribute even when creating a singular object.

src/pkgcore/ebuild/repo_objs.py
```diff
@@ -170,6 +171,217 @@ def __init__(self, metadata_xml, manifest):
...
+
+            subprojects = []
+            for sp in p.findall('subproject'):
+                subprojects.append(Subproject(
+                    ref=sp.get('ref'),
+                    inherit_members=sp.get('inherit-members', '') == '1',
+                    projects_xml=self))
+            kwargs['subprojects'] = tuple(subprojects)
```
radhermit: Same thing here as above.

src/pkgcore/ebuild/repo_objs.py
```diff
@@ -170,6 +171,217 @@ def __init__(self, metadata_xml, manifest):
...
+    """
+
+    __slots__ = (
+        "__weakref__", "_projects", "_source"
+    )
+
+    def __init__(self, source):
+        self._source = source
+
+    @property
```
radhermit: Probably another jitted attr here since that's the behavior you're approximating.

src/pkgcore/ebuild/repo_objs.py
```diff
@@ -170,6 +171,217 @@ def __init__(self, metadata_xml, manifest):
...
+    def leads(self):
+        """Project lead(s), if any."""
+        return tuple(m for m in self.members if m.is_lead)
+
+    @property
+    def recursive_members(self):
+        """All project members, including members inherited from subprojects."""
+        # collect all subprojects from which to inherit, recursively
```
radhermit: Are the levels of projects always restricted to be at most three by policy? Currently this code appears to cap out at project -> subproject -> subproject which isn't quite recursive.

mgorny: You sure about that? I don't see any reason why the code wouldn't go on forever, as long as new projects are appended to `subprojects` list.

radhermit: Nope, looks like I'm incorrect. However I'm still wondering, how many levels is the deepest we currently get?

src/pkgcore/ebuild/repo_objs.py
mgorny: I managed to make jitattr work but this one still fails.

mgorny: this one = dirproxy

src/pkgcore/ebuild/repo_objs.py
mgorny: Done. Though I'm not sure if it actually increases readability here.

src/pkgcore/ebuild/repo_objs.py
mgorny: I have no clue and I don't feel like counting ;-P.

src/pkgcore/ebuild/repo_objs.py
mgorny: Made it non-fatal and added extra tests. It kinda sucks for subproject indirection API but should be doable.

mgorny:
All applied, I think.

radhermit:
Thanks, merged with some simplifications/minor changes. Note that you'll need that latest snakeoil for fixes to klass.GetAttrProxy to support jitted attrs and klass.DirProxy to support classes with __slots__.

PR #266, opened 2019-08-03
Fix package.environment getter

mgorny:


radhermit:
Thanks, merged with some minor changes.

PR #267, opened 2019-08-03
repo_objs: Include type in Maintainer for metadata.xml

mgorny:
Introduce a 'maint_type' (since 'type' is a builtin) attribute
in Maintainer, and use to indicate maintainer type from metadata.xml.


radhermit:
Thanks, merged.

PR #276, opened 2020-01-02
Sync/sqfs cleanup

ferringb:
1) add comment explaining why http is used for sqfs syncer
2) Fix up the http syncer to handle 304 more properly, and also throw in some defensive code while I'm in here for type handling since the original code looks wrong.
3) also tweak the http syncer to not care about protocol upper/lower case.

### DIFF
--- a/data/config/repos.conf
+++ b/data/config/repos.conf
@@ -4,6 +4,8 @@ main-repo = gentoo
 [gentoo]
 location = /var/db/repos/gentoo
 repo-type = sqfs-v1
+# distfiles.gentoo.org certs aren't actually valid, defeating the purpose of https; thus
+# forcing http.
 sync-uri = http://distfiles.gentoo.org/snapshots/squashfs/gentoo-current.lzo.sqfs
 sync-type = sqfs
 
--- a/src/pkgcore/sync/http.py
+++ b/src/pkgcore/sync/http.py
@@ -6,10 +6,11 @@ import ssl
 import sys
 import urllib.request
 
-from snakeoil.fileutils import AtomicWriteFile
+from snakeoil.fileutils import AtomicWriteFile, readfile_ascii
 from snakeoil.osutils import pjoin
 
 from pkgcore.sync import base
+from pkgcore.log import logger
 
 
 class http_syncer(base.Syncer):
@@ -36,22 +37,12 @@ class http_syncer(base.Syncer):
 
         if not force:
             # use cached ETag to check if updates exist
-            previous_etag = None
-            try:
-                with open(etag_path, 'r') as f:
-                    previous_etag = f.read()
-            except FileNotFoundError:
-                pass
+            previous_etag = readfile_ascii(etag_path, none_on_missing=True)
             if previous_etag:
                 headers['If-None-Match'] = previous_etag
 
             # use cached modification timestamp to check if updates exist
-            previous_modified = None
-            try:
-                with open(modified_path, 'r') as f:
-                    previous_modified = f.read()
-            except FileNotFoundError:
-                pass
+            previous_modified = readfile_ascii(modified_path, none_on_missing=True)
             if previous_modified:
                 headers['If-Modified-Since'] = previous_modified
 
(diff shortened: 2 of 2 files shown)


ferringb:
Ok, I now understand why y'all used http for the protocol- distfiles.gentoo.org certs aren't actually valid.

I'm not sure whether the correct response there is to throw a chair or point folks at using redirects, but it is what it is.

In light of that, I'm modifying the https commit to explain why an insecure approach is taken.

radhermit:
> Ok, I now understand why y'all used http for the protocol- distfiles.gentoo.org certs aren't actually valid.
> 
> I'm not sure whether the correct response there is to throw a chair or point folks at using redirects, but it is what it is.
> 
> In light of that, I'm modifying the https commit to explain why an insecure approach is taken.

Yep, I vaguely recall hitting that and iirc infra guys generally provide the excuse "all distributed files have checksums and/or gpg signatures so https isn't necessary." Personally, if that were my opinion I'd also disable https:// support entirely for those endpoints as well, but that's just me. :smile:

ferringb:
Yeah, it's abjectly cracked, but arguing w/ them (including about the fact that http allows history replay attacks to force old versions) isn't something I expect they'll care about.

@mgorny - being the sqfs backend is now the default for pkgcore installations (which is how I caught this), is there a saner location you know of for distribution of the squashfs image?  I may have interest in implementing the gpg validation, but if the content isn't over https- securing this is pointless due to the underlying replay attacks.

mgorny:
Hmm, I thought we were actually limiting distfiles.gentoo.org to servers with HTTPS cert. Maybe something got broken recently?

radhermit:
> Hmm, I thought we were actually limiting distfiles.gentoo.org to servers with HTTPS cert. Maybe something got broken recently?

It would be great if they had support. I'm not sure what you mean by "recently" but I think this has been the case for at least a year now since that's about when I started using squashfs syncing/repo support in pkgcore.

ferringb:
@mgorny easy test:
```
brian@hrair:~$ openssl s_client -connect distfiles.gentoo.org:443 | openssl verify
depth=2 C = US, ST = New Jersey, L = Jersey City, O = The USERTRUST Network, CN = USERTrust RSA Certification Authority
verify return:1
depth=1 C = US, ST = MI, L = Ann Arbor, O = Internet2, OU = InCommon, CN = InCommon RSA Server CA
verify return:1
depth=0 C = US, postalCode = 97331, ST = OR, L = Corvallis, street = 121 Valley Library, O = Oregon State University, OU = CASS, CN = *.osuosl.org
verify return:1
C = US, postalCode = 97331, ST = OR, L = Corvallis, street = 121 Valley Library, O = Oregon State University, OU = CASS, CN = *.osuosl.org
error 20 at 0 depth lookup: unable to get local issuer certificate
error stdin: verification failed
```
more obvious report:
```
brian@hrair:~$ curl -v https://distfiles.gentoo.org/
*   Trying 2605:bc80:3010::134...
* TCP_NODELAY set
* Connected to distfiles.gentoo.org (2605:bc80:3010::134) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; postalCode=97331; ST=OR; L=Corvallis; street=121 Valley Library; O=Oregon State University; OU=CASS; CN=*.osuosl.org
*  start date: Jul 24 00:00:00 2018 GMT
*  expire date: Jul 23 23:59:59 2020 GMT
*  subjectAltName does not match distfiles.gentoo.org
* SSL: no alternative certificate subject name matches target host name 'distfiles.gentoo.org'
* stopped the pause stream!
* Closing connection 0
* TLSv1.2 (OUT), TLS alert, Client hello (1):
curl: (51) SSL: no alternative certificate subject name matches target host name 'distfiles.gentoo.org'
```
^^^ core thing, the CN/SAN don't cover distfiles.gentoo.org- they're just a *.osuosl.org .  I doubt that that the distfiles mirrors are allowed to have cert's with distfiles.gentoo.org, so this setup- in terms of security design- can't ever have valid https/PKI.

The usual approach folks use here is to have distfiles.gentoo.org be a bouncer- a redirecter- which allows PKI validation that we're talking w/ distfiles.gentoo.org, and then validation that we're talking w/ the host that they redirected us at.

Either way- this is broken.

mgorny:
I've pinged Robin about it and he said he's got a short-term solution planned.

radhermit:
Thanks, merged with a fix for checking missing etag/modified headers during http sync on top.

Also, just to note it's easy to catch such breakage if you run the real syncing tests that have to be explicitly enabled for obvious reasons. For example, run something similar to ``./setup.py test -a "-m net --network"`` or ``pytest -m net --network`` in the root dir of the repo to run all tests marked as real network tests.

PR #279, opened 2020-04-10
Support <stabilize-allarches/> in metadata.xml

mgorny:


mgorny:
Ping.

radhermit:
Thanks, merged.

mgorny:
ETOOLATE. I've already reinvented the wheel, and added `restrict` support. Any suggestion how to go about that in pkgcore? I suppose it would either mean unmaking metadata.xml 'shared' or storing the data along with restrict and filtering on package level.

radhermit:
Yes, the latter is how I'd do it. For each element that you wanted to support restrict attributes for you'd probably have to store a mapping or tuple of restrictions/values. Then the package class that provides an external interface to those values for each package would have to check the restrictions when the attribute was accessed and collapse to the proper values (so probably a jitted attr).

PR #284, opened 2020-08-29
Prep work in restrictions and atom area

ferringb:
To resolve the issues with combinatorial explosion of transitive use deps, the underlying restriction framework is going to have to be refactored.  Simply put the notion of 'force_True' equivalent operations on a repository were ideas that didn't pan out due to EAPI going in a way I didn't predict, and the CNF/DNF explosions that have followed are derived from that.

To fix that underlying problem, the notion of "mutable configurable packages" in a repository is going to have to be separate out- likely into another layer (domain, resolver plan, etc) that holds the changes in flight.

To get there- to evaluate if that is right- will require cleanup up restrictions and introducing more formalized visitation.  Currently we have the notion of visiting values and 'package'; 'package' is where the combinatorial explosion occurs for transitive use apps, and it's not going to possible to change that.  Fact is I'm not sure we should, anyways.

What's needed is to add another higher level of visitation that exposes CNF/DNF while exposing the configurable relationships to the visiter.  The point there is to make the visiter deal w/ a higher level API, but one that allows it to introspect/slice/dice/whatever on it's own, allowing it to operate more efficiently than a forced normalized form.

That's the end goal, and this PR is prep work- cleanup and removal of anything particularly ugly I left years ago.  The commits detail this in full, but relevant points:

1) src/atom.c should be dead at this point since it's py2k only.  Either way, it had dead load statements which I pruned since I noticed them while doing other cleanup.
2) values.ContainmentMatch2 is what everything uses, and was to replace values.ContainmentMatch a decade ago.  I forgot to do that cleanup/replacement.  At this point I've updated all of pkgcore code and just aliased the usages together since pkgcheck has some awareness. A proper warning of deprecation seems warranted, although I'm kind of fine w/ being lazy on that front.
3) MInor rename w/in atom to give a more clear property name for 'no_usedeps'; at first glance one can't tell if it's a boolean indicating "it has no usedeps", or if it's something that does a render.  I've left an alias in place for compatiability, and have renamed it to 'get_atom_without_use_deps' for clarity.

### DIFF
--- a/src/pkgcore/ebuild/atom.py
+++ b/src/pkgcore/ebuild/atom.py
@@ -16,7 +16,7 @@ from snakeoil.demandload import demand_compile_regexp
 from ..restrictions import boolean, packages, restriction, values
 from ..restrictions.packages import AndRestriction as PkgAndRestriction
 from ..restrictions.packages import Conditional
-from ..restrictions.values import ContainmentMatch2
+from ..restrictions.values import ContainmentMatch
 from . import cpv, errors, restricts
 
 # namespace compatibility...
@@ -649,7 +649,7 @@ class transitive_use_atom(atom):
 
     @staticmethod
     def _mk_conditional(flag, payload, negate=False):
-        return Conditional('use', ContainmentMatch2(flag, negate=negate), payload)
+        return Conditional('use', ContainmentMatch(flag, negate=negate), payload)
 
     def _recurse_transitive_use_conds(self, atom_str, forced_use, varied):
         if not varied:
--- a/src/pkgcore/ebuild/atom.py
+++ b/src/pkgcore/ebuild/atom.py
@@ -446,8 +446,10 @@ class atom(boolean.AndRestriction, metaclass=klass.generic_equality):
 
         return cmp(self.repo_id, other.repo_id)
 
+    no_usedeps = klass.alias_attr("get_atom_without_use_deps")
+
     @property
-    def no_usedeps(self):
+    def get_atom_without_use_deps(self):
         """Return atom object stripped of USE dependencies."""
         if not self.use:
             return self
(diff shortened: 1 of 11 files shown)


radhermit:
> 1. src/atom.c should be dead at this point since it's py2k only.  Either way, it had dead load statements which I pruned since I noticed them while doing other cleanup.

Personally I don't think the C extensions are worth keeping around anymore. For my part, I didn't want to maintain both the C extension side and an equivalent pure python side going into the py3 era. If anything I'd rather rewrite in an entirely different language. :laughing:

> 2. values.ContainmentMatch2 is what everything uses, and was to replace values.ContainmentMatch a decade ago.  I forgot to do that cleanup/replacement.  At this point I've updated all of pkgcore code and just aliased the usages together since pkgcheck has some awareness. A proper warning of deprecation seems warranted, although I'm kind of fine w/ being lazy on that front.

I doubt the few 3rd party uses of the pkgcore API use this, it's probably fine to yank without a deprecation period.

> 3. MInor rename w/in atom to give a more clear property name for 'no_usedeps'; at first glance one can't tell if it's a boolean indicating "it has no usedeps", or if it's something that does a render.  I've left an alias in place for compatiability, and have renamed it to 'get_atom_without_use_deps' for clarity.

IIRC, it's something I needed in pkgcheck to generate equivalent atom objs without usedeps attached.

ferringb:
Revisiting this:
> Values.ContainmentMatch2
This can be trivially renamed/cleaned up.  I've got PRS to do that, which I'll push here shortly.  From there, pkgcheck will have to cut a version that has a dep for a higher version, but that's that.

ferringb:
This collapses it down to some minor cleanup of the repo.itermatch related internals, the rename of ContainmentMatch2 -> ContainmentMatch, and the aforementioned atom class cleanup.

Once that lands, I'll cut pkgcheck PRs to clean things up.

mgorny:
Rebasing for a CI run, then I'll merge it.

thesamesam:
> This collapses it down to some minor cleanup of the repo.itermatch related internals, the rename of ContainmentMatch2 -> ContainmentMatch, and the aforementioned atom class cleanup.
> 
> Once that lands, I'll cut pkgcheck PRs to clean things up.

Fancy doing those? 😄

PR #286, opened 2020-08-29
Less magic package _get_attr machinery

ferringb:
Back in the old days, how this was written was the only way to achieve certain performance requirements (back when rotational was the norm, and this seriously mattered).  Also, decorators weren't available at the time, so that was a bit of a different era..  Times changed, and the functional style in use here doesn't make contributions easy for others, let alone easy to read.

The basic `_get_attr` approach can't be ripped out in a single shot due to pkgcheck knowledge of it, and due to the metadata package's `release_cached_data` need to know all attributes that were rendered, and remove them.  Said another way, this still has problems.

What this PR implements instead is refactoring the code into a pattern where `snakeoil.klass.jit_attr` can eventually be dropped in and the internal bookkeeping of `_get_attr` removed.

The commit messages have further details for each, and are worth a read if anything is unclear.

One final point- there were a few attributes that weren't being rendered and stored to the package instance.  IE, space was allocated for it- things like `tracked_attributes`.  In those cases I've converted the code so that it's a property- rendered each access- since access is infrequent for common use cases, but the memory overhead does exist.  Minor, but no point in wasting in this case.

### DIFF
--- a/src/pkgcore/ebuild/cpv.py
+++ b/src/pkgcore/ebuild/cpv.py
@@ -242,7 +242,6 @@ class CPV(base.base):
     :ivar revision: str revision
     :ivar versioned_atom: atom matching this exact version
     :ivar unversioned_atom: atom matching all versions of this package
-    :cvar _get_attr: mapping of attr:callable to generate attributes on the fly
     """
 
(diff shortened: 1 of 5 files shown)


radhermit:
Merged with some minor updates due to code changes, thanks.

Issue #177, opened 2016-02-12
EbuildProcessor.is_alive(): does some crazy /proc/*/status lookups


mgorny:
Long story short, it uses snakeoil's `process.is_running()`. That function does some random lookup for `/proc/*/status`, which is far from portable and even further from reliable. Well, I'm not sure but it looks like this may cause pretty funky results if ebuild processor dies and the PID is reused by another process.

I'm pretty sure pkgcore would be much happier doing proper process management with `SIGCHLD`, `os.wait()` and other nice stuff.

ferringb:
this was resolved via the work started in 7a71d5d, the issue should be closed.

PR #287, opened 2020-08-29
update EbuildProcessor.is_alive to use waitpid

ferringb:
The original is_alive property was written to deal w/ very old python
versions that had issues with non-main threads doing signal handling,
thus this code.

That issue no longer exists, or I'm recalling the 'why' wrong and the code was just daft from day one.

Either way, just use an os.waitpid() instead. #177 is further context.

The surrounding PID handling still is attrocious and in needed of cleanup,
but that's a later date.

Note this splits the notion of 'is_alive' into "is_alive" (a process exists), and "is_responsive".  The code
actually cares a bit in this regard- it probably shouldn't- but it does.  Splitting the states to be clearer
should make it easier to rewrite the processor around async/coroutine notions at a later date, hence
why I'm making that change along w/ the waitpid related cleanup.

### DIFF
--- a/src/pkgcore/ebuild/processor.py
+++ b/src/pkgcore/ebuild/processor.py
@@ -134,7 +134,7 @@ def release_ebuild_processor(ebp):
     assert ebp not in inactive_ebp_list
     # We can't reuse processors that use custom fd mappings or are locked on
     # release for one reason or another.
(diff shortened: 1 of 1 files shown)


radhermit:
Merged, thanks.

PR #300, opened 2021-03-14
Support <maintainer proxied=""/>

mgorny:
Per the GLEP 67 update approved today.


radhermit:
Thanks, merged with a couple minor changes.

PR #308, opened 2021-05-09
ebuild.processor: Ensure that signal handlers are set only once

mgorny:
Prevent the signal.signal() calls from being made more than once.  This
was changed in 533f1edd70054a5479ee85719d3cbef0d15627fd when the calls
were moved to EbuildProcessor's __init__ method.  However, it seems
to cause 'pmaint regen' to hang randomly which keeps blocking Gentoo
git repository mirrors.


mgorny:
So apparently this doesn't fix the issue completely. I couldn't reproduce the hangs on martin but they still happen on jacamar :-(.

PR #309, opened 2021-05-15
pytest: Delay loading pkgcore modules until fixtures are used

mgorny:
Delay loading pkgcore modules until the EbuildRepo-based fixtures are
actually used.  This prevents the pkgcore signal handlers from being
enabled on all packages using pytest while keeping the old behavior
of setting them upon import in packages using pkgcore directly.

### DIFF
--- a/src/pkgcore/ebuild/processor.py
+++ b/src/pkgcore/ebuild/processor.py
@@ -268,6 +268,9 @@ def chuck_KeyboardInterrupt(*args):
     raise KeyboardInterrupt("ctrl+c encountered")
 
 
+signal.signal(signal.SIGINT, chuck_KeyboardInterrupt)
+
+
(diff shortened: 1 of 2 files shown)


mgorny:
FYI, I've confirmed that pkgcore's, pkgcheck's and pytest's tests work with this one (pytest was previously broken by these signal handlers). I've applied it to CI's sources to and I'm currently monitoring for pmaint hanging.

radhermit:
I don't like this workaround since it avoids the issue that allows other consumers to use threading with ebd instances and still run into the same issues.

radhermit:
Closing since the underlying threading issues should be fixed via https://github.com/pkgcore/pkgcore/commit/fcc786e9cef88a16e83932ef5cbc1090c01d4f32.

PR #310, opened 2021-05-15
EAPI 8

mgorny:
- [x] Selective fetch restriction
- [x] Install-time CBUILD dependencies
- [ ] <del>Variant of \|\| ( ) with defined runtime behaviour</del>
- [x] Pass --datarootdir to configure
- [x] Pass --disable-static to configure
- [x] Accumulate PROPERTIES & RESTRICT over eclasses and ebuilds
- [ ] <del>RESTRICT value for network-restricted tests</del>
- [x] dosym -r to create symlinks relative to link location
- [x] Second optional argument for usev
- [x] Less strict naming rules for files in updates directory
- [x] Bash 5.0
- [x] Default src_prepare accepts only file names in PATCHES
- [x] More consistent insopts/exeopts
- [x] Ban useq, hasq, and hasv functions
- [x] unpack: Remove support for 7-Zip, RAR, and LHA


mgorny:
It's ready now. `RESTRICT=network-test` will most likely be replaced by retroactive `PROPERTIES=network-test`, and it will not be strictly necessary for anything.

ebd/eapi/8/phase.bash
```diff
@@ -5,3 +5,15 @@ usev() {
 	fi
 	return 1
 }
+
```
radhermit, line 8: PKGCORE_BANNED_FUNCS should be used instead of explicitly defining stub functions.

src/pkgcore/ebuild/pkg_updates.py
```diff
@@ -8,22 +8,19 @@
...
-        match = valid_updates_re.match(filename)
+        match = eapi.options.update_regex.match(filename)
         if match is not None:
-            files.append(((match.group(2), match.group(1)), filename))
+            files.append(filename)
         else:
             logger.error(f'incorrectly named update file: {filename!r}')
-    files.sort(key=itemgetter(0))
-    return [x[1] for x in files]
+    return files
```
radhermit, line 20: I don't think sorting should be removed here even though the spec doesn't specify it's required since the order of iteration probably affects pkgcheck output among other things. If EAPI 8 doesn't care what the files are named, then use a basic lexical sort.

ebd/eapi/8/phase.bash
mgorny: This works for useq but not hasq, hasv. If I create `ebd/eapi/8/depend.bash`, it does not seem to be used at all. If I add them in `global.bash` or `phase.bash`, pkgcore complains:

```
EAPI 8: undefined banned func: 'hasq'

src/pkgcore/ebuild/pkg_updates.py
mgorny: Done.

ebd/eapi/8/phase.bash
radhermit: > This works for useq but not hasq, hasv. If I create `ebd/eapi/8/depend.bash`, it does not seem to be used at all. If I add them in `global.bash` or `phase.bash`, pkgcore complains:
> 
> ```
> EAPI 8: undefined banned func: 'hasq'
> ```

That's fixed now in git so all function names should be able to be placed in the PKGCORE_BANNED_FUNCS array.

mgorny: Ok, thanks.

mgorny:
Updated per comments.

mgorny:
(and rebased)

radhermit:
Merged, thanks for your work. I've briefly glanced over all the changes, but since EAPI 8 is light we can fix any other potential issues later.

PR #315, opened 2021-06-15
[DO NOT MERGE, HELP NEEDED] pkg_uptodate experiment

mgorny:
@radhermit, this is what I've been able to hack so far. However, it fails with:

```pytb
Sandboxed process killed by signal: Broken pipe
Traceback (most recent call last):
  File "/home/mgorny/git/pkgcore/.venv/lib/python3.10/site-packages/pkgcore/ebuild/ebd.py", line 471, in run_generic_phase
    if not ebd.run_phase(phase, env, tmpdir=tmpdir, sandbox=sandbox,
  File "/home/mgorny/git/pkgcore/.venv/lib/python3.10/site-packages/pkgcore/ebuild/processor.py", line 428, in run_phase
    if not self.send_env(env, tmpdir=tmpdir):
  File "/home/mgorny/git/pkgcore/.venv/lib/python3.10/site-packages/pkgcore/ebuild/processor.py", line 725, in send_env
    fileutils.write_file(path, 'wb', data.encode())
  File "/home/mgorny/git/pkgcore/.venv/lib/python3.10/site-packages/snakeoil/fileutils.py", line 43, in write_file
    f = open(path, mode, encoding=encoding)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/portage/app-misc/hello-2.10-r1/temp/ebd-env-transfer'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/mgorny/git/pkgcore/test.py", line 13, in <module>
    ret = op._generic_phase('uptodate', True, True)
  File "/home/mgorny/git/pkgcore/.venv/lib/python3.10/site-packages/pkgcore/ebuild/ebd.py", line 323, in _generic_phase
    return run_generic_phase(
  File "/home/mgorny/git/pkgcore/.venv/lib/python3.10/site-packages/pkgcore/ebuild/ebd.py", line 497, in run_generic_phase
    raise format.GenericBuildError(
pkgcore.operations.format.GenericBuildError: failed build operation: Executing phase uptodate: Caught exception: [Errno 2] No such file or directory: '/tmp/portage/app-misc/hello-2.10-r1/temp/ebd-env-transfer'
```

The addition to hello ebuild is:
```bash
pkg_uptodate() {
	einfo "Yep, boss!"
	return 0
}
```

### DIFF
--- a/ebd/ebuild.bash
+++ b/ebd/ebuild.bash
@@ -391,7 +391,7 @@ __execute_phases() {
 		PKGCORE_SUPPRESS_BASHRCS=false
 
 		case ${EBUILD_PHASE} in
-			nofetch|pretend)
+			nofetch|pretend|uptodate)
 				PKGCORE_SUPPRESS_BASHRCS=true
 				__init_environ
 
--- a/src/pkgcore/ebuild/ebuild_src.py
+++ b/src/pkgcore/ebuild/ebuild_src.py
@@ -343,7 +343,8 @@ class base(metadata.package):
     @property
     def mandatory_phases(self):
         return frozenset(
-            chain(self.defined_phases, self.eapi.default_phases))
+            chain(self.defined_phases, self.eapi.default_phases,
+                  ['uptodate']))
 
     @property
     def live(self):
--- a/test.py
+++ b/test.py
@@ -1,0 +1,17 @@
+#!/usr/bin/env python
+
+import pkgcore.config
+import pkgcore.ebuild.atom
+import pkgcore.ebuild.ebd
+
+c = pkgcore.config.load_config(location='/etc/portage')
+d = c.get_default('domain')
+r = d.installed_repos[0]
+pkg = r.match(pkgcore.ebuild.atom.atom('app-misc/hello'))[0]
+
+op = pkgcore.ebuild.ebd.misc_operations(d, pkg)
+ret = op._generic_phase('uptodate', True, True)
+print(ret)
+
+import IPython
+IPython.embed()


mgorny:
How I'm testing: I've created a venv with `--system-site-packages`, and i'm installing `pip install .` inside, then running `python test.py`.

ferringb:
@mgorny I'm assuming this was an attempt to have another phase added?  If so, update this w/ details/links (an issue link would be great) and I can take a look.

mgorny:
Rough idea was to have a custom phase, and be able to call it externally from `smart-live-rebuild`. The function would return zero if the live ebuild is up-to-date (i.e. doesn't need rebuilding), or non-zero if it needs rebuilding.

ferringb:
@mgorny any qualms if I close this out?  I don't track gentoo-dev/mls, so if this is still an occasional discussion, pardon.

That said, I'd like to get our open PRs and issues fully under prioritized control, and close out the stale stuff.

Flowdalic:
FTR, Sam pointed this issue out after I asked him if `openpgp-keys-gentoo-developers-9999` could become git based, so that I could run `smart-live-rebuild -f openpgp-keys-gentoo-developers` before `emerge --sync` to avoid sync fails due to `openpgp-keys-gentoo-developers-9999` not being up-to-date.

A `pkg_uptodate()` helper phase would also be a viable solution. But I wonder if this wouldn't also require a standardized way to query the installation time of a package. In case of `openpgp-keys-gentoo-developers-9999` you could probably fairly trivial obtain the last-modified value, but you would need to compare it with sometime. For example, the emerge time of the package. I also note that there is probably a small window where this could produce false negatives. The more robust solution would be `openpgp-keys-gentoo-developers-9999`  to also install the the active-devs.gpg file and rely on mtime preservation (if that works) to compare the mtime of the file with the value of the last-modified header.

PR #316, opened 2021-06-17
Start pkg_* phases in a dedicated empty directory

### LINKED ISSUE #313: EAPI 8: please implement initial dir update for non-src_* phases
EAPI 8 has the following new rule:

>  If no initial working directory is mandated, then for EAPIs listed in table 9.1 as having an empty directory, it must be set to a dedicated directory that is empty at the start of the function and may be read-only.
>
> https://projects.gentoo.org/pms/8/pms.html#x1-860009.1.1

This is not critical but I think you'll be able to figure out where to put it faster than I would ;-).

mgorny:
Start pkg_* phases in a dedicated empty directory that's recreated
for every phase invoked.  This is required by EAPI 8 but since earlier
EAPIs allow any starting directory, it's fine to do it for all EAPIs.

Fixes #313

### DIFF
--- a/ebd/ebuild.bash
+++ b/ebd/ebuild.bash
@@ -348,6 +348,7 @@ __run_ebuild_phase() {
 	source "${PKGCORE_EBD_PATH}"/.generated/libs/${EAPI}/$1 \
 		|| die "failed sourcing EAPI ${EAPI} $1 phase scope lib"
 
+	cd "${PKGCORE_EMPTYDIR}" || die "unable to cd into ${PKGCORE_EMPTYDIR}"
 	__qa_run_function_if_exists __phase_pre_$1
 	__qa_run_function_if_exists pre_$1
 
--- a/src/pkgcore/ebuild/ebd.py
+++ b/src/pkgcore/ebuild/ebd.py
@@ -211,7 +211,8 @@ class ebd:
         for x, y in (("T", "temp"),
                      ("WORKDIR", "work"),
                      ("D", "image"),
-                     ("HOME", "homedir")):
+                     ("HOME", "homedir"),
+                     ("PKGCORE_EMPTYDIR", "empty")):
             self.env[x] = normpath(pjoin(self.builddir, y))
         self.env["D"] += self.eapi.options.trailing_slash
         self.env["PORTAGE_LOGFILE"] = normpath(pjoin(self.env["T"], "build.log"))
@@ -313,6 +314,10 @@ class ebd:
                 return True
             if 'selinux' not in self.features and 'suidctl' not in self.features:
                 return True
+
+        shutil.rmtree(self.env["PKGCORE_EMPTYDIR"], ignore_errors=True)
+        os.mkdir(self.env["PKGCORE_EMPTYDIR"])
+
         userpriv = self.userpriv and userpriv
         sandbox = self.sandbox and sandbox
         self._set_per_phase_env(phase, self.env)


src/pkgcore/ebuild/ebd.py
```diff
@@ -211,7 +211,8 @@ def _set_op_vars(self, tmp_offset):
         for x, y in (("T", "temp"),
                      ("WORKDIR", "work"),
                      ("D", "image"),
-                     ("HOME", "homedir")):
+                     ("HOME", "homedir"),
+                     ("PKGCORE_EMPTYDIR", "empty")):
```
ulm, line 215: Wouldn't this better be named `__PKGCORE_EMPTYDIR`, since it's an internal variable?

ulm:
Ping. Any progress here?

I'd really like to add Pkgcore to the EAPI 8 row of this table:
https://wiki.gentoo.org/wiki/Project:Package_Manager_Specification#Implementation_in_package_managers

ebd/ebuild.bash
```diff
@@ -348,6 +348,7 @@ __run_ebuild_phase() {
 	source "${PKGCORE_EBD_PATH}"/.generated/libs/${EAPI}/$1 \
 		|| die "failed sourcing EAPI ${EAPI} $1 phase scope lib"
 
+	cd "${PKGCORE_EMPTYDIR}" || die "unable to cd into ${PKGCORE_EMPTYDIR}"
```
radhermit, line 351: Isn't this supposed to only be done for pkg_* phases? This runs it for every phase.

src/pkgcore/ebuild/ebd.py
radhermit: I don't think the base tmpdir is exported in any fashion so using a separate variable is required for now in that sense to support it cleanly rather than recreating it on the ebd side using a different var and appending ``../empty``.

Wrt the ``__`` prefix, pkgcore doesn't use the same approach as portage and automatically filters all ``PKGCORE_`` prefixed-vars from the exported env, so it should be fine as it is now.

radhermit:
> Ping. Any progress here?
> 
> I'd really like to add Pkgcore to the EAPI 8 row of this table:
> https://wiki.gentoo.org/wiki/Project:Package_Manager_Specification#Implementation_in_package_managers

From my perspective, it is functionally EAPI 8 compliant for how 99.9999%+ people use pkgcore so I'd have already added it to the table.

ebd/ebuild.bash
mgorny: Do you see any harm in doing it unconditionally, then having phases with different definition change workdir? It seemed cleaner to me this way.

radhermit: > Do you see any harm in doing it unconditionally, then having phases with different definition change workdir? It seemed cleaner to me this way.

Not off the top of my head... this would be one of those merge and see if it breaks the few people who use pkgcore for build/merging kind of commits that I was alluding to. :laughing:

ulm:
> From my perspective, it is functionally EAPI 8 compliant for how 99.9999%+ people use pkgcore so I'd have already added it to the table.

Sorry, but we add package managers to that table when they support all features of a new EAPI, not a subset of features.

radhermit:
> > From my perspective, it is functionally EAPI 8 compliant for how 99.9999%+ people use pkgcore so I'd have already added it to the table.
> 
> Sorry, but we add package managers to that table when they support all features of a new EAPI, not a subset of features.

Technically then you should remove more lines from that table since iirc there are a few other things that have never been finished for a couple EAPIs.

PR #323, opened 2021-09-01
ebuild.eclass: Add support for recursively grabbing @PROVIDES [WIP]

mgorny:
TODO: tests

### DIFF
--- a/src/pkgcore/ebuild/eclass.py
+++ b/src/pkgcore/ebuild/eclass.py
@@ -319,7 +319,8 @@ _eclass_blocks_re = re.compile(
 class EclassDoc(AttrDict):
     """Support parsing eclass docs for a given eclass path."""
 
-    def __init__(self, path, /, *, sourced=False):
+    def __init__(self, path, /, *, sourced=False, repo=None):
+        self.repo = repo
         self.mtime = os.path.getmtime(path)
 
         # parse eclass doc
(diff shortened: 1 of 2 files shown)


src/pkgcore/ebuild/eclass.py
```diff
@@ -409,6 +410,22 @@ def live(self):
...
+        if self.repo is None:
+            raise ValueError('EclassDoc.provides can only be used if the class has been initialized with repo')
+        out = OrderedSet()
+        to_process = OrderedSet(self.raw_provides)
+        while to_process:
+            out.add(next_eclass := to_process.pop())
+            if (next_eclass_inst := self.repo.eclass_cache.get_eclass(next_eclass)) is None:
+                logger.warning(f"'@PROVIDES:' eclass {next_eclass} not found")
+                continue
+            next_eclass_doc = EclassDoc(next_eclass_inst.path)
```
radhermit, line 425: At a glance, it appears this needlessly re-parses core eclasses that are inherited by others a lot of additional times. This is why I recommended wrapping this support back at the repo level to avoid this, but if you want to do it this way then you'd make EclassDoc instances immutable and use caching via snakeoil caching/lru/etc.

Beyond that, it's a property that is called per package in InheritsCheck for pkgcheck. This means the level of unnecessary parsing just expanded exponentially when using pkgcheck.

radhermit: Reading a bit more, it looks like set handling should avoid re-parsing internally per-eclass, but I think the point still stands that this is run per pkg by pkgcheck with no cache used to alleviate that at the repo level or internally.

radhermit: Yet another option (and probably the easiest) I'd recommend is doing this during EclassDoc initialization and dumping the data in an attr (e.g. '_provides') that will be saved to the pkgcheck eclass cache objects similar to how _source_eclass() works. Then provide access to that attr via a property so the majority of the runtime work (parsing) isn't duplicated when accessing it in this fashion.

PR #325, opened 2021-09-03
ebuild.eclass_cache: Implement caching getter for EclassDocs

mgorny:
(as part of the solution to problems in #323)

I've tried using WeakValueDictionary but it seems to have discarded the cache almost immediately which made it suck a lot.


mgorny:
@radhermit confirmed my feelings that this is ugly.

PR #326, opened 2021-09-05
ebuild.eclass: Add ABI_VERSION attribute to EclassDoc

mgorny:
Add ABI_VERSION attribute that will be increased whenever EclassDoc's
internal ABI changes in incompatible ways.  This will be used to drive
pkgcheck's cache version, and therefore make it possible to make
incompatible changes inside pkgcore without having to keep manually
bumping the cache version in pkgcheck.


mgorny:
@radhermit, what do you think about this idea?

radhermit:
> @radhermit, what do you think about this idea?

Sounds fine.

mgorny:
Merged

Issue #318, opened 2021-08-04
pmerge sets libdir to lib64 if no --libdir is given in the ebuild


oz123:
I'm not sure if it's a real issue. `emerge` has a much larger and more extensive `econf` function which can handle
ebuilds that don't specify it.

Specifically, I encountered this issue when compiling postgresql with pmerge, and I was surprised to see a different result with emerge.

A more detail discussion can also be found in https://bugs.gentoo.org/806362.

mgorny:
From Gentoo bug: https://projects.gentoo.org/pms/8/pms.html#x1-128002r2

I'll try to take a look at it today.

I presume that you're actually using pkgcore to install stuff? That's cool because I don't, and it's great to have someone actually testing that ;-).

ulm:
Looks like a simple typo:
https://806362.bugs.gentoo.org/attachment.cgi?id=730080

oz123:
@mgorny yes, i really like pkgcore it's solving dependencies much faster than portage. 
One critic I have is that error messages are not as friendly as with emerge.

PR #328, opened 2021-09-23
ebd_ipc: Fix unpack functionality broken via refactoring.

ferringb:
Rough tracing takes this back to logic being added to allow
for fetchers that have differing distdirs; the problem is
there is a break in the encapsulation.

Basically the fetcher has to return the location of the
content on disk, but just returns a boolean indicating success.

Being the invoking operation (setup usually) wasn't getting
updated for where the distfiles where (becaused the `verified_files`
attribute wasn't updated), unpack functionality was broken.

This represented via files being downloaded, but not being symlinked
into the spoof DISTDIR; thus no unpack.

### DIFF
--- a/src/pkgcore/ebuild/ebd.py
+++ b/src/pkgcore/ebuild/ebd.py
@@ -679,7 +679,10 @@ class buildable(ebd, setup_mixin, format.build):
         # fetch distfiles
         if not self.verified_files:
             ops = self.domain.pkg_operations(self.pkg, observer=self.observer)
-            ops.fetch()
+            if ops.fetch():
+                # this break encapsulation and should be refactored.  Trace
+                # f35f2 and 6561eac for where this was refactored.
+                self.verified_files = ops.verified_files
 
         # symlink them into builddir
         if self.verified_files:


mgorny:
I'm confused. How can I reproduce this breakage?

ferringb:
> I'm confused. How can I reproduce this breakage?

Either invoke pmerge, or invoke pebuild.  *All* EBD unpack invocations are broken- the 'shadow $DISTDIR' code wasn't being ran, so ebuilds were being given a $DISTDIR w/ no content in it.

Easy repro:
`pebuild dev-util/diffball clean unpack`

As to the fix; as I mentioned in the code comments, it's encapsulation breaking; this is due to the encapsulation *being broken* by the revisions I referred to.  That code should be refactored, but restoring the ability to use pkgcore as a PM is a first step.

mgorny:
Then I'm afraid I can't reproduce this, and I honestly doubt @radhermit would have missed such a huge problem given that he is (was?) using pkgcore as his primary PM. I mean, I've just tried a few random packages and `pmerge` is working just fine.

I'm not saying I don't trust that you know what you're doing. However, I'd like to understand what's happening, why I can't reproduce this and why tests don't catch this.

ferringb:
No need for the extra verbosity (I obviously can make mistakes and I'm
aware of it). I can repro this on two machines, If you can't, then there is
a variable we are missing that is the cause.

Can you post your python version + features variable please?

I'll use that as a known good, and trace down what differs for my hosts and
y'all's.

Thanka-

mgorny:
> Can you post your python version + features variable please?

Python 3.10.0rc2 (but I've just tried 3.8.12 too)

I'll give you the whole `pconfig dump`: [pconfig-dump.txt](https://github.com/pkgcore/pkgcore/files/7232108/pconfig-dump.txt)

Maybe @radhermit can have some suggestions, I'm using Portage-first system here, his configs are probably more kosher.

thesamesam:
FWIW, we've had a few other using pmerge for various bits with pmerge. That's how we found #318. I also believe radhermit was/is until very recently using it and was able to install e.g. Firefox.

src/pkgcore/ebuild/ebd.py
```diff
@@ -679,7 +679,10 @@ def _setup_distfiles(self):
         # fetch distfiles
         if not self.verified_files:
             ops = self.domain.pkg_operations(self.pkg, observer=self.observer)
-            ops.fetch()
+            if ops.fetch():
+                # this break encapsulation and should be refactored.  Trace
+                # f35f2 and 6561eac for where this was refactored.
+                self.verified_files = ops.verified_files
```
ferringb, line 685: First, note that nothing in this method actually updates `self.verified_files` without my change.  That means the conditional that follows lien 688- will not fire unless something invoked by `ops.fetch()` mutates this instance.

Take a look at https://github.com/pkgcore/pkgcore/commit/f35f2c#diff-9e1b9f482d50cb7dabc3871cae3e57f9366c304bb8eaf9f0fae77a576ad91992L685-L690 ; you'll see the previous code broke encapsulation and transfered the `verified_files` mutable from the fetcher directly, and that that commit tried to feed it back through the operations API.

That broke things in a different way, which lead to https://github.com/pkgcore/pkgcore/commit/6561eacc4225f0b602d1612ee02b2b9d20d7a93d , which is where we're at now.

I honestly do not see anyway that this codepath could work and I suspect if you trace what I'm pointing, you'll reach a similar conclusion.

https://github.com/pkgcore/pkgcore/blob/master/src/pkgcore/ebuild/ebd.py#L670-L676 also shows the DISTDIR setup; it's basically the samething as the 'saviour' days, it uses a shadow $DISTDIR, and the host configuration can't really override that.

The only way this pathway would work is if `verified_files` was pre-populated and being passed in, or if the build init was doing something I'm not finding.  See https://github.com/pkgcore/pkgcore/blob/master/src/pkgcore/ebuild/ebd.py#L605-L608 for what I'm referring to.

ferringb:
I honestly don't see how the unpack functionality could be working for any ebuild that relies on `$DISTDIR` (those that use `$PORTAGE_ACTUAL_DISTDIR` would sidestep the bug).  See https://github.com/ferringb/pkgcore/issues/2 for an example failure (note: I'm using my fork as my own staging ground, rather than actually forking the issue tracking).  That failure is due to /var/tmp/portage/dev-util/diffball-1.0.1-r2/distdir' never being populated.

@mgorny my repro for you may be incorrect.  `pebuild =dev-util/diffball-1.0.1-r2 clean unpack` should trigger it; the -9999 build falls back to git, thus uses `$PORTAGE_ACTUAL_DISTDIR`.

My suggestion is folks look through https://github.com/pkgcore/pkgcore/pull/328/files#r716277302 ; I'm guessing y'all won't see how this function could work, and will see that how I'm changing it, restores the encapsulation violation that previously made it work.

thesamesam:
```
$ pebuild =dev-util/diffball-1.0.1-r2 clean unpack
WARNING:pkgcore:repos.conf: parsing '/etc/portage/repos.conf/gentoo.conf': overriding DEFAULT section
executing phase unpack
 * ERROR: dev-util/diffball-1.0.1-r2::gentoo failed (unpack phase):
 *   unpack: exitcode 1: nonexistent file: 'diffball-1.0.1.tar.gz'
 *
 * Call stack:
 *    isolated-functions.bash, line 333:  called __phase_src_unpack
 *                     global, line  62:  called unpack 'diffball-1.0.1.tar.gz' 'diffball-1.0.1-bug_543310_stack_buffer_overflows.patch' 'diffball-1.0.1-bug_708736_cseek_xz_reset_avail_in_out.patch'
 *                 src_unpack, line 145:  called __ebd_ipc_cmd 'unpack' '' 'diffball-1.0.1.tar.gz' 'diffball-1.0.1-bug_543310_stack_buffer_overflows.patch' 'diffball-1.0.1-bug_708736_cseek_xz_reset_avail_in_out.patch'
 *     ebuild-daemon-lib.bash, line  91:  called __ipc_exit '1' 'nonexistent file: 'diffball-1.0.1.tar.gz''
 *     ebuild-daemon-lib.bash, line  75:  called die
 *
 * If you need support, post the topmost build error, and the call stack if relevant.
 *
 * Working directory: '/var/tmp/portage/dev-util/diffball-1.0.1-r2/work'
```

vs
```
$ pmerge =dev-util/diffball-1.0.1-r2
WARNING:pkgcore:repos.conf: parsing '/etc/portage/repos.conf/gentoo.conf': overriding DEFAULT section
 * Resolving...

 * Running sanity checks...


Processing 1 of 1: dev-util/diffball-1.0.1-r2::gentoo

3 files required-
building dev-util/diffball-1.0.1-r2
>>> Unpacking diffball-1.0.1.tar.gz to /var/tmp/portage/dev-util/diffball-1.0.1-r2/work
 * unpack: skipping unrecognized file format: 'diffball-1.0.1-bug_543310_stack_buffer_overflows.patch'
 * unpack: skipping unrecognized file format: 'diffball-1.0.1-bug_708736_cseek_xz_reset_avail_in_out.patch'
 * Applying diffball-1.0.1-bug_543310_stack_buffer_overflows.patch...
 * Applying diffball-1.0.1-bug_708736_cseek_xz_reset_avail_in_out.patch...
 * Running eautoreconf in '/var/tmp/portage/dev-util/diffball-1.0.1-r2/work/diffball-1.0.1' ...
 * Running 'libtoolize --install --copy --force --automake' ...                                                                                                                                        [ ok ]
 * Running 'aclocal' ...                                                                                                                                                                               [ ok ]
 * This package has a configure.in file which has long been deprecated.  Please
 * update it to use configure.ac instead as newer versions of autotools will die
 * when it finds this file.  See https://bugs.gentoo.org/426262 for details.
 * Running 'autoconf --force' ...                                                                                                                                                                      [ ok ]
 * Running 'autoheader' ...                                                                                                                                                                            [ ok ]
 * Running 'automake --add-missing --copy --foreign --force-missing' ...
^C
```

ferringb:
@thesamesam thanks for that, that confirms the bug I'm reporting in the underlying internals.  `pebuild` tends to invoke the operation directly- hence the `clean unbuild` arguments.  `pmerge` invokes the final step.

On a rough guess, the operation step digraph that `pmerge` is triggering is leading to `verified_files` being populated and passed in as I mentioned in https://github.com/pkgcore/pkgcore/pull/328#discussion_r716277302 .  `pebuild` is lower level and invoking just the phases necessary, which is what's exposing it.

Either way, I reiterate; there is zero way the _setup_distfiles method is working as is required/intended.  `pmerge` not exposing the bug is an annoying bit of luck.

radhermit:
Skimming through the comments without reading everything, ferringb is generally correct here... I know since I have an unpushed fix with the exact same code change since I noticed ``pebuild`` broken at some point a few weeks ago when I was testing some rust code for pkgcraft. :laughing:

In other words, feel free to merge the change.

ferringb:
@radhermit Appreciate the clarification/confirmation my read was right.

Upshot, I'm still paying for lax test standards back in the day. ;)

radhermit:
> @radhermit Appreciate the clarification/confirmation my read was right.

No problem, thanks for taking the time to look into things as well since I unfortunately got distracted and never pushed my version of the fix. :laughing:

radhermit:
With regards to mentions about my usage of pmerge, I'm down to a single ~13 year old machine using it (or portage) which will probably get decommissioned in the next year so I'm not the best person to look to for reproducing pkg*-related issues.

I mostly just look at the code on occasion for ideas when trying to figure out how to redesign support using idiomatic rust these days.

PR #329, opened 2021-09-27
Improver resolver UX for failures

ferringb:
I'm basically convinced the resolver (and the underlying configurable abstraction/layer) needs a rewrite- it's been a problem since '08- but to get there, and to not annoy users, the UX needs to at least be more verbose in indicating what is and isn't a failure.  

Thus this PR: it adds color and fixes indentation.

Colors are:
* green == normal
* yellow == unsolved
* red == failure point.

Before:
image

After:
image

Folks may not consider this all that important, but if you're working on resolver code, having non-shit failure output is kind of a requirement.


mgorny:
Thanks!

PR #330, opened 2021-09-27
ebd: update EAPI 8 --disable-static rule per PMS update

mgorny:
Update the EAPI 8 --disable-static to be emitted when either
--disable-static or --enable-static occurs in configure --help output.
This is necessary to handle libtool packages correctly, and this is how
it was initially implemented in Portage.  PMS is being updated to match.

Bug: https://bugs.gentoo.org/814368


ulm (commented):
LGTM

PR #331, opened 2021-12-11
provide access to inherited package.deprecated and use_expand_desc

CyberTailor:
Accumulate `deprecated` and `use_expand_desc` attributes not only from the repo but also from its masters.

(see conversation in pkgcore/pkgcheck#334)


mgorny:
Thank you!

PR #333, opened 2021-12-16
processor: allow _ prefix for Bash variables

thesamesam:
With /etc/portage/env/foo containing e.g. "__PORTAGE_FOO=0", one would get:
```
Sandboxed process killed by signal: Broken pipe
!!! caught exception building sys-libs/glibc-2.34-r3: failed build operation:
Executing phase setup: Caught exception: "_PORTAGE_CONFCACHE: bash doesn't allow digits as the first char"
pmerge: error: failed build operation: Executing phase setup: Caught exception:
"_PORTAGE_CONFCACHE: bash doesn't allow digits as the first char"
```

It's fine for variables to begin with an underscore as well as A-z.

Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/src/pkgcore/ebuild/processor.py
+++ b/src/pkgcore/ebuild/processor.py
@@ -691,7 +691,7 @@ class EbuildProcessor:
         for key, val in sorted(env_dict.items()):
             if key in self._readonly_vars:
                 continue
-            if not key[0].isalpha():
+            if not key[0].isalpha() and not key.startswith("_"):
                 raise KeyError(f"{key}: bash doesn't allow digits as the first char")
             if not isinstance(val, (str, list, tuple)):
                 raise ValueError(


thesamesam:
CI [failure](https://github.com/pkgcore/pkgcore/runs/4545150715?check_suite_focus=true#step:9:162) seems to be a missing dep in snakeoil?

thesamesam:
so that I don't forget:
```
[21:07:39]  <@sam_> mgorny: https://github.com/pkgcore/pkgcore/pull/333 pls
[21:15:00]  <@ulm> sam_: is isalpha right? I thought that only ASCII letters are allowed
[21:15:09]  <@ulm> not all of unicode
[21:15:13]  <@sam_> let's check
[21:16:08]  <@sam_> https://docs.python.org/3/library/stdtypes.html#str.isalpha we can add 'and .isascii()'
[21:24:52]  <@mgorny> sam_: maybe .isidentifier() + .isascii()? the rules seem to roughly match bash
```

src/pkgcore/ebuild/processor.py
```diff
@@ -691,7 +691,7 @@ def _generate_env_str(self, env_dict):
         for key, val in sorted(env_dict.items()):
             if key in self._readonly_vars:
                 continue
-            if not key[0].isalpha():
+            if not key[0].isalpha() and not key.startswith("_"):
                 raise KeyError(f"{key}: bash doesn't allow digits as the first char")
```
ferringb, line 695: error message needs updating.

ferringb:
First glance, change looks fine sans the error message being updated.

src/pkgcore/ebuild/processor.py
thesamesam: Thanks. I hadn't got back to this as I got distracted & wanted to make the change mgorny suggested because it was unclear if we should be allowing _more_ than just alpha + _, but I don't mind just going with your change.

PR #336, opened 2022-03-17
Rename @ECLASS-VARIABLE to @ECLASS_VARIABLE

ulm:


src/pkgcore/ebuild/eclass.py
```diff
@@ -235,6 +235,29 @@ def _supported_eapis(self, block, tag, lineno):
...
+            '@PRE_INHERIT': ('pre_inherit', False, self._tag_bool, False),
+            '@USER_VARIABLE': ('user_variable', False, self._tag_bool, False),
+            '@OUTPUT_VARIABLE': ('output_variable', False, self._tag_bool, False),
+            '@DESCRIPTION:': ('description', True, self._tag_multiline_str, None),
+        }
+        super().__init__(tags)
+
+
+# For backwards compatibility, can be removed after 2022-12-31
+class EclassVarBlockCompat(ParseEclassDoc):
```
mgorny, line 260: Hmm, no chance you could use `EclassVarBlock` as the base class and just override `tag`?

ulm: That's why I had asked in #-qa. :smile:
```
<@mgorny> i think duplicating is fine
<@mgorny> or even better
```

`@ECLASS-VARIABLE` is also used in `tags`, so would have to override that too (how?).

ulm: Sorry, but I fear that my python skills aren't sufficient to produce a more elegant solution. Especially, not sure how the child class would access `tags`, which is defined inside the `__init__` method of its parent.

tests/ebuild/test_eclass.py
```diff
@@ -64,13 +64,13 @@ def make_eclass(name, provides=None):
...
 
-# @ECLASS-VARIABLE: _FOO_INTERNAL_ECLASS_VAR
+# @ECLASS_VARIABLE: _FOO_INTERNAL_ECLASS_VAR
 # @INTERNAL
 # @DEFAULT_UNSET
 # @DESCRIPTION:
 # Internal variable.
 
-# @ECLASS-VARIABLE: FOO_PUBLIC_ECLASS_VAR
+# @ECLASS_VARIABLE: FOO_PUBLIC_ECLASS_VAR
```
mgorny, line 73: How about keeping the old syntax in one of them for better test coverage? Either that or adding a dedicated test for the compat logic.

thesamesam:
conceptual ack

tests/ebuild/test_eclass.py
ulm: I've added `FOO_ANOTHER_ECLASS_VAR` with the old syntax.

mgorny (approved):
Thanks for doing this! Maybe I'll experiment with trying to simplify the code later.

PR #337, opened 2022-03-24
ebuild.eclass: Warn about deprecated @ECLASS-VARIABLE tag

ulm:


mgorny:
Can you update the test to make sure the warning gets emitted? Grep for `pytest.warns()`.

ulm:
> Can you update the test to make sure the warning gets emitted? Grep for `pytest.warns()`.

```
pkgcore $ grep -ri warns
pkgcore $ 
```

mgorny:
Sorry, I meant grep in pytest docs: https://docs.pytest.org/en/6.2.x/warnings.html#warns

Basically, the test you modified to check if `@ECLASS-VARIABLE` works should emit your new warning. I want you to capture it via pytest to make sure it's emitted. If you don't want to play with that, I can do it.

PR #339, opened 2022-04-20
ebuild.ebuild_src: close resources when possible in eapi()

laumann:
With the following setup:

    $ export PYTHONWARNINGS=d,i::ImportWarning
    $ export PYTHONTRACEMALLOC=1
    $ pkgcheck scan ./app-editors/vim/vim-8.2.0814-r100.ebuild

produces a lot of warnings of this kind:

    /usr/lib/python3.9/site-packages/pkgcore/package/base.py:97: ResourceWarning: unclosed file <_io.TextIOWrapper name='./app-editors/vim/vim-9999.ebuild' mode='r' encoding='utf8'>
      val = functor(self)
    Object allocated at (most recent call last):
      File "/usr/lib/python3.9/site-packages/snakeoil/_fileutils.py", lineno 78
        handle = open(mypath, mode, encoding=encoding)

For the one ebuild above there are ~150 such warnings. The majority of
the calls appear to be the call to fileutils.readlines_utf8() that
produces a file handle that is not properly closed.

Signed-off-by: Thomas Bracht Laumann Jespersen <t@laumann.xyz>

### DIFF
--- a/src/pkgcore/ebuild/ebuild_src.py
+++ b/src/pkgcore/ebuild/ebuild_src.py
@@ -252,6 +252,7 @@ class base(metadata.package):
             if (mo := _EAPI_str_regex.match(line)) and (eapi_str := mo.group('EAPI')):
                 eapi = _EAPI_regex.match(line).group('EAPI')
             break
+        i.close()
         try:
             return get_eapi(eapi)
         except ValueError as e:


laumann:
I really would've preferred to use a `with` block, but that would've required moving more code around. I'm open to doing that, this was just the smallest change that made the majority of the resource warnings go away.

Here's the full log from the `pkgcheck scan` run: https://0x0.st/oc77.ebuild%3B

src/pkgcore/ebuild/ebuild_src.py
```diff
@@ -252,6 +252,8 @@ def eapi(self):
             if (mo := _EAPI_str_regex.match(line)) and (eapi_str := mo.group('EAPI')):
                 eapi = _EAPI_regex.match(line).group('EAPI')
             break
+        if hasattr(i, "close"):
```
mgorny, line 255: Is there a case that it wouldn't have the `close` method? It feels weird to check for that.

laumann: It does feel weird. The `i` variable is either

```python
i = fileutils.readlines_utf8(ebuild.path)
```
or
```python
i = (x.strip() for x in ebuild.text_fileobj())
```
(I've only seen the first variant so far).

In the latter case, does `ebuild.text_fileobj()` need to be closed?

mgorny: Probably.

laumann: OK, reworked to call close() for the fileutls.readlines_utf8(), but I've skipped the other variant, simply because I haven't seen it executed yet :-)

laumann:
> Just close unconditionally, we know that the other call will return a `file` object.

Done.

PR #342, opened 2022-05-23
Sync metadata.xsd from proj/xml-schema repository

ulm:


ulm:
Note that this bundled file is only used as a fallback for pkgcheck if the repo doesn't provide it.
See pkgcore/pkgcheck@1bbcf79082d986bb8fae7d6a66912e53192033e7.

ulm:
Not sure what your policy on these updates is. So, if you think that these should be squashed into one commit then go ahead.

arthurzam (approved):

Issue #345, opened 2022-05-24
Improve eclass doc generation


arthurzam:
The final target is a good generator used for `app-doc/eclass-manpages` and for `devmanual`.
There is a need for adding missing sections, and cleaning a lot of things!
ulm requested the output as DocBook output. There are existing RST to DocBook convertors.

## Resources:

- [eclass → manpage](https://github.com/mgorny/eclass-to-manpage)
- [manpage → html for devmanual](https://gitweb.gentoo.org/proj/devmanual.git/tree/bin/gen-eclass-html.sh#n145)
- [rst → dockbook using docutils](https://github.com/Abstrys/rst2db/blob/master/abstrys/docutils_ext/docbook_writer.py)
- [RST docutils spec](https://docutils.sourceforge.io/docs/user/rst/quickref.html#block-quotes)

## Bugs:

- https://bugs.gentoo.org/846503
- https://bugs.gentoo.org/773271
- https://github.com/pkgcore/pkgcore/issues/297

## Project name:
pkgcore's devmanual eclass doc pages generator (`pdedpg` for short)

ulm:
> The final target is a good generator used for `app-doc/eclass-manpages` and for `devmanual`. There is a need for adding missing sections, and cleaning a lot of things! ulm requested the output as DocBook output. There are existing RST to DocBook convertors.

Not DocBook, but DevBook XML. See [human readable](https://devmanual.gentoo.org/appendices/devbook-guide/index.html) and [formal](https://gitweb.gentoo.org/proj/devmanual.git/plain/devbook.rnc) specs. It's a pretty simple format, especially when limited to eclass documentation, where there are no complicated elements like tables or figures.

If this should prove to be too difficult, then HTML would also be acceptable (after all, that's what we have now).

arthurzam:
First of all, thank you @ulm for pointing me into the correct format (no idea how I managed to mistake docbook and devbook).
I'm now in the process of writing a translator from `rst` to `devbook` (your links to explanation and spec were spot on). It progresses not bad, and the remaining shouldn't be too hard.

I think if this finishes as intended, we will have a good `eclassdoc → {rst,manpage,devbook}` convertor. Then I think we can decide that eclassdoc is rst format (not so far from current state, and then devs can better format them). Finally, I guess I can do a traversal of the eclasses to improve the format of eclassdoc, to better confirm. Main issues I find:

- missing blank line between bullet list items.
[...]

PR #346, opened 2022-06-04
pmaint eclass: new devbook format

arthurzam:
This is a PR for the new implementation of devbook (the format for devmanual) format for `pmaint`'s docs generator. This is intended to replace the current eclass2manpage and manpage2html usage.
You can see the current output at my [devspace](https://dev.gentoo.org/~arthurzam/devmanual/eclass-reference/index.html)
I have also tested the output of the manpage, and it looks at least the same (sometimes even better) then current once.
I have done some local edits for the devmanual makefile, so it can use this new facility (currently it uses a local hack of a symlink to gentoo repo inside src, so I still can't upstream it).

I now assume the all eclassdocs are RST formatted (except the normal special tags of course). I will traverse all eclasses and fix/update the docs, to match the new format (there are currently some last places where those are wrongly formatted). This work will be done, and sent to ML to await comments and ideas. I don't plan to merge this PR until later.

Closes: https://github.com/pkgcore/pkgcore/issues/345


src/pkgcore/ebuild/eapi.py
```diff
@@ -715,3 +715,5 @@ def _combine_dicts(*mappings):
     )),
     ebd_env_options=eapi7._ebd_env_options,
 )
+
+latest_eapi = eapi8
```
ferringb, line 719: I'm not a huge fan of 'latest' here; pkgcore's original intent was to support varying EAPI's, including developmental.  Is a developmental one 'latest' ?

TL;DR: can we call this `latest_pms_eapi` ?

ferringb: Hell, we could have `pms_eapis = [obj1, obj2, obj3, ...]` and allow folks to just do `pms_eapis[-1]` to get latest.  Thoughts?

arthurzam: I agree. This was a draft PR, more as a "WIP" branch for the future. You can ignore this PR from any review for the near future.

Flowdalic:
What's the status of this? @thesamesam mentioned you said that the code is done and ist just needs the docs porting? We recently had [the case where eclass-to-manpage failed](https://github.com/gentoo/gentoo/commit/7c9bacebd7254491a72e633b523feacfec72d5ec), therefore having proper eclass verification as part of the CI would be very nice. And now the question is if we wait for this or maybe add the eclass-to-manpage awk script to the CI.

PR #353, opened 2022-08-06
ebuild/atom.py: atom.no_usedeps should include subslot

arthurzam:
We don't have a lot of users of the property no_usedeps for atom, so it was missed that we were ignoring the subslot value in the return. By including it, we are fixing an issue that was found with `VirtualKeywordsUpdate` in `virtual/podofo-build`.

Tested over at `::gentoo` repo, and it indeed fixed it for `virtual/podofo-build`.

Resolves: https://github.com/pkgcore/pkgcheck/issues/411

### DIFF
--- a/tests/ebuild/test_atom.py
+++ b/tests/ebuild/test_atom.py
@@ -548,3 +548,23 @@ class Test_atom(test.TestRestriction):
         self.assertFalse(self.kls("=dev-util/diffball-1").is_simple)
         self.assertFalse(self.kls("dev-util/diffball[x]").is_simple)
         self.assertFalse(self.kls("dev-util/diffball[x?]").is_simple)
+
+    def test_get_atom_without_use_deps(self):
+        for original, wanted in (
+            ("<dev-util/diffball-2", "<dev-util/diffball-2"),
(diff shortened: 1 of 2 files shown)


tests/ebuild/test_atom.py
```diff
@@ -548,3 +548,23 @@ def test_unversioned(self):
         self.assertFalse(self.kls("=dev-util/diffball-1").is_simple)
         self.assertFalse(self.kls("dev-util/diffball[x]").is_simple)
         self.assertFalse(self.kls("dev-util/diffball[x?]").is_simple)
+
+    def test_get_atom_without_use_deps(self):
+        for original, wanted in (
```
mgorny, line 553: `pytest.mark.parametrize` perhaps?

arthurzam: For now problematic, as I can't use fixtures as this test class inherits from `TestCase`.

In a separate branch I'm refactoring those test classes to not use it, so let's use ugly `for` for this PR?

mgorny: Sure. I thought this was post-refactoring already ;-).

PR #355, opened 2022-08-08
ci: add pypy3

thesamesam:
Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,16 +13,16 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.10', '3.11', 'pypy-3.10']
+        python-version: ['3.10', '3.11', '3.12', 'pypy-3.10']
         deps: [newest-deps]
         experimental: [false]
         include:
           - os: ubuntu-latest
-            python-version: '3.11'
+            python-version: '3.12'
             deps: minimal-deps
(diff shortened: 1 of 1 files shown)


ferringb:
PR needs refreshing; as to pypy, I'm not opposed although I stopped tracking the status of it probably a decade back.  More to the point: is this something that we should prevent a fix from being merged over?  IE: if it builds on cpy but not pypy, is pypy worth blocking from?

arthurzam:
> PR needs refreshing; as to pypy, I'm not opposed although I stopped tracking the status of it probably a decade back. More to the point: is this something that we should prevent a fix from being merged over? IE: if it builds on cpy but not pypy, is pypy worth blocking from?

I'm not sure I understand all you wrote, so I will add myself information.

We meant here pypy3, which is quite well supported in Gentoo, and is an interesting check if pkgcore stack would be faster on pypy3 instead of cpython.

The fact the test suite is failing, means we need to investigate the reason - is it a real breakage or some "esoteric" feature. My gut feeling is it relates to namespace walking.

thesamesam:
I think @ferringb was asking if future changes should be blocked once we add this to CI if they fail only on pypy. I would say yes, because it's likely to imply something odd is being done. But I'm fine with just leaving this for now (this isn't mergeable until we fix the existing pypy issues anyway).

PR #362, opened 2022-09-02
patom: new command

arthurzam:
Requested by @mgorny
The args and docs are taken from `atom`, but the format will have different names.
I still have various extra arguments I want to implement, but this is a start.

- [x] Add tests
- [ ] Implement `--compare`

### DIFF
--- a/src/pkgcore/scripts/patom.py
+++ b/src/pkgcore/scripts/patom.py
@@ -1,0 +1,105 @@
+"""atom parsing utility"""
+
+import re
+from functools import partial
+
+from ..ebuild.atom import atom as atom_cls
+from ..ebuild.errors import MalformedAtom
+from ..util.commandline import ArgumentParser
+
+
+def atom(value: str) -> atom_cls:
+    try:
+        return atom_cls(value)
+    except MalformedAtom as exc:
+        # try to add an operator in case we got a version without op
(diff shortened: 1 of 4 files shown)


src/pkgcore/scripts/patom.py
```diff
@@ -0,0 +1,78 @@
...
+        Conversion specifiers start with a ``%`` symbol and are followed by
+        either ``{`` or ``[``.  Next is the name of the field to expand,
+        followed by a matching ``}`` or ``]``.
+
+        The difference between ``{`` and ``[`` is that the latter is only
+        printed if the field referred is set, while the former prints
+        ``<unset>`` in that case.
+
+        The following fields are supported, which are inline with the
+        variables from ebuild(5).
```
mgorny, line 24: …but they're not? ;-)

src/pkgcore/scripts/patom.py
```diff
@@ -0,0 +1,78 @@
...
+            revision ``-r0`` is not printed.
+
+        REVISION
+            The ebuild revision.
+
+        SLOT
+            The package slot.
+
+        SUBSLOT
+            The package sub slot.
```
mgorny, line 46: You may want to clarify how these work when subslot exists and not.

PR #372, opened 2022-10-21
New custom flit backend

arthurzam:
Still needs much testing, and comparison that all files exists with similar contents!


Makefile
```diff
@@ -0,0 +1,15 @@
...
+man html:
+	$(SPHINX_BUILD) -a -b $@ doc build/sphinx/$@
+
+.PHONY: sdist wheel
+sdist wheel:
+	$(PYTHON) -m build --$@
+
+.PHONY: clean
+clean:
+	$(RM) -r build/sphinx doc/api
```
mgorny, line 15: Don't you need to define `RM`? Also, you want `-f` or it'll fail when dirs don't exist.

doc/conf.py
```diff
@@ -227,25 +226,32 @@
...
 
 # -- Options for manual page output --------------------------------------------
+generate_man(os.path.abspath('..'), '', 'pkgcore')
+generate_html(os.path.abspath('..'), os.path.abspath('../src'), 'pkgcore')
 
-bin_path = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))), 'bin')
-scripts = os.listdir(bin_path)
+try:
+    import tomllib
+except ImportError:
```
mgorny, line 234: Let's do it via `if sys.version_info`, that's better handled by random tools.

pyproject.toml
```diff
@@ -1,11 +1,87 @@
...
+    "License :: OSI Approved :: BSD License",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+]
+dynamic = ["version"]
+
+dependencies = [
+	"snakeoil~=0.10.0",
```
mgorny, line 32: Inconsistent indentation.

mgorny (commented):
Just took a quick look, didn't review fully.

Makefile
mgorny: Hmm, I didn't know that. I suppose it's fine.

pyproject.toml
```diff
@@ -1,11 +1,85 @@
...
+[tool.flit.sdist]
+include = [
+	"tox.ini", ".coveragerc", "Makefile", "py_build.py",
+	"NEWS.rst", "doc", "tests", "examples", "bin",
+	"build/sphinx/man/*.1", "build/sphinx/man/*.5",
+]
+exclude = [
+	".github/", ".gitignore",
+	"doc/api/", "doc/generated/",
+]
```
mgorny, line 74: I personally prefer these one per line but up to you.

pyproject.toml
```diff
@@ -1,11 +1,85 @@
...
+
+[project.entry-points."pytest11"]
```
mgorny, line 76: BTW does `pytest11` need to be quoted here? It looks weird but maybe it does.

mgorny (approved):
I think it's as good as it can get. `pycodestyle` will probably have a few complaints still.

arthurzam:
@mgorny many thanks. Currently the lessons learned from this process are enough for me to do myself `pkgdev`, but for `pkgcheck` I will do some more thinking and open future discussion.

PR #374, opened 2022-11-01
setup.py: reorder data files

arthurzam:
Store all data files (files that should be installed into `/usr` outside python lib location) in one main "data" directory, with direct subdirs matching the once installed on system. This is preparation for moving to flit backend.

This separate PR was created to make it easier to review https://github.com/pkgcore/pkgcore/pull/372 (split the mass renames from flit changes).

The only "practical" changes are only at root directory (`.gitignore`, `MANIFEST.in`, `setup.py`).


setup.py
```diff
@@ -239,14 +241,7 @@ def write_pkgcore_lookup_configs(python_base, install_prefix, injected_bin_path=
     author_email='radhermit@gmail.com',
     entry_points={'pytest11': ['pkgcore = pkgcore.pytest.plugin']},
     data_files=list(chain(
```
mgorny, line 243: I think you don't need `chain()` anymore.

Issue #376, opened 2022-11-15
Plugin cache is stale after install/upgrade


floppym:
When I run pmaint regen after installing/upgrading sys-apps/pkgcore, an error message is logged:
```
ERROR:pkgcore:Cannot write cache for /usr/lib/python3.11/site-packages/pkgcore/plugins/plugincache: [Errno 13] Permission denied: '/usr/lib/python3.11/site-packages/pkgcore/plugins/.update.plugincache'. Try running pplugincache.
```
It seems strange that it would try to regenerate the plugin cache immediately given that the plugincache file is generated as part of the installation process.

radhermit:
Just to note for historical context, in my opinion pkgcore's plugin system is another misfeature where effort to remove it or restructure it entirely would probably be the best path towards a maintainable future. I wiped its traces from pkgcheck when the rewrite occurred, but never had the energy to restructure pkgcore in the same fashion. If something like it still existed in pkgcheck, then (for example) you'd have to keep a hardcoded list of keywords and checks manually updated instead of the semi-dynamic system that's used now where those lists are generated at runtime and/or during the install process.

IIRC, the issue with reworking the plugin system for pkgcore is that it's entangled with the config setup which is also over-engineered so I entirely understand continuing to "brush it under the rug". :laughing: :

arthurzam:
@radhermit Thanks for the input.

This plugin system does a lot of issues and hard to plan with it, so I want to get rid of it. I'm thinking to maybe drop it totally, or make it constant (similar to `_objects.py` in pkgcheck).

PR #377, opened 2022-11-15
plugin: also consider plugincache file's mtime

arthurzam:
Also use plugincache file's mtime to determine if the plugin cache need update. Add a grace period of 1 second for the mtime check to avoid cases of unfortunate mtime changes in the file system unpack time.

Resolves: https://github.com/pkgcore/pkgcore/issues/376


arthurzam:
Doesn't work and won't work - bad idea!

PR #382, opened 2022-12-24
Reformat the repository with black 22.12.0

ferringb:
Additionally, add a .git-blame-ignore-revs so git blame doesn't attribute all of this to myself. Once that file is in there I'll go through the history and add a few more since snakeoil history is long, and code has shuffled enough the original commit documentation isn't easily reachable.

Note: if folks don't like the formatting of this, the PR can be punted. Black can also be tuned as needed, this is just me getting the ball rolling.


thesamesam:
I like it. We did it for Portage and I don't regret it at all, and nobody else has voiced any complaints.

arthurzam (approved):

Issue #296, opened 2021-02-09
OSError: [Errno 24] Too many open files ; when regenerating a large repo


mgorny:
```pytb
$ time pmaint regen -t 16 --force --use-local-desc --pkg-desc-index -vvv ros-overlay
Traceback (most recent call last):
  File "/usr/lib/python-exec/python3.9/pmaint", line 4, in <module>
    scripts.run(basename(__file__))
  File "/usr/lib/python3.9/site-packages/pkgcore/scripts/__init__.py", line 32, in run
    sys.exit(tool())
  File "/usr/lib/python3.9/site-packages/snakeoil/cli/tool.py", line 79, in __call__
    ret = self.main()
  File "/usr/lib/python3.9/site-packages/snakeoil/cli/tool.py", line 185, in main
    self.handle_exec_exception(e)
  File "/usr/lib/python3.9/site-packages/snakeoil/cli/tool.py", line 170, in main
    exitstatus = func(self.options, self.out, self.err)
  File "/usr/lib/python3.9/site-packages/pkgcore/scripts/pmaint.py", line 314, in regen_main
    ret.append(repo.operations.regen_cache(
  File "/usr/lib/python3.9/site-packages/pkgcore/operations/__init__.py", line 69, in _recast_exception_decorator
    return functor(*args, **kwds)
  File "/usr/lib/python3.9/site-packages/pkgcore/operations/repo.py", line 252, in _cmd_api_regen_cache
    pkgs = frozenset(pkg.cpvstr for pkg in self.repo)
  File "/usr/lib/python3.9/site-packages/pkgcore/operations/repo.py", line 252, in <genexpr>
    pkgs = frozenset(pkg.cpvstr for pkg in self.repo)
  File "/usr/lib/python3.9/site-packages/pkgcore/repository/prototype.py", line 297, in _internal_match
    for pkg in self._internal_gen_candidates(candidates, **kwargs):
  File "/usr/lib/python3.9/site-packages/pkgcore/repository/prototype.py", line 294, in _internal_gen_candidates
    yield from sorter(pkg_filter(pkgs))
  File "/usr/lib/python3.9/site-packages/pkgcore/ebuild/repository.py", line 557, in _pkg_filter
    pkg.data
  File "/usr/lib/python3.9/site-packages/pkgcore/package/base.py", line 97, in dynamic_getattr_dict
    val = functor(self)
  File "/usr/lib/python3.9/site-packages/pkgcore/package/metadata.py", line 50, in data
[...]

PR #383, opened 2022-12-24
Fix secondary exception from when an EBD fails to start.

ferringb:
Further in this `__init__`, `self.pid` is initialized to the process that is started.  If that fails- an exception is thrown- then self.pid doesn't exist, which in turn breaks the assumptions of `shutdown_processor()` and `.is_alive` (both of which are designed to work if an init partially failed).

Finally, add protections to ensure a failed spawn doesn't leak file descriptors.  Python won't reap raw FD's, so we need to handle this gracefully.

This secondary exception is visible in pkgcore/pkgcore#296

### DIFF
--- a/src/pkgcore/ebuild/processor.py
+++ b/src/pkgcore/ebuild/processor.py
@@ -324,6 +324,7 @@ class EbuildProcessor:
         self._eclass_caching = False
         self._outstanding_expects = []
         self._metadata_paths = None
+        self.pid = None
 
         spawn_opts = {'umask': 0o002}
         if self.userpriv:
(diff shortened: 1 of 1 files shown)


src/pkgcore/ebuild/processor.py
```diff
@@ -371,25 +372,37 @@ def __init__(self, userpriv, sandbox, fd_pipes=None):
...
-
-        # force each ebd instance to be a process group leader so everything
-        # can be easily terminated
-        self.pid = spawn_func(
-            [spawn.BASH_BINARY, self.ebd, "daemonize"],
-            fd_pipes=ebd_pipes, returnpid=True, env=env, pgid=0, **spawn_opts)[0]
-
-        os.close(cread)
-        os.close(dwrite)
+        try:
```
ferringb, line 377: a second set of eyes on the logic for cleanup/suppression would be appreciated- this is extremely edge case to trigger, I'm fixing it because I can see it in the code, not because I've managed to trigger it.

arthurzam: Thinking about all the possible flows, it looks like you are super careful and catch all the things. I was checking for an exception in each statement, and I couldn't find a miss.

arthurzam (approved):
Great job! Will merge in a bit.

Issue #393, opened 2023-01-15
parsing error: token -nginx_modules_http_* is not a valid use flag


mgorny:
It seems that pkgcore doesn't handle `-*` entries in `package.use` correctly, e.g.:

```
WARNING:pkgcore:'/etc/portage/package.use/local.conf', line 2: parsing error: token -nginx_modules_http_* is not a valid use flag
WARNING:pkgcore:'/etc/portage/package.use/local.conf', line 3: parsing error: token -apache2_modules_* is not a valid use flag
WARNING:pkgcore:'/etc/portage/package.use/local.conf', line 6: parsing error: token -l10n_* is not a valid use flag
```

The respective lines are:

```
*/* NGINX_MODULES_HTTP: -* gzip limit_conn limit_req rewrite spdy uwsgi
*/* APACHE2_MODULES: -* actions alias deflate filter mime mime_magic negotiation rewrite socache_shmcb unixd
[…]
*/* L10N: -* pl
```

thesamesam:
Yeah, I get the same for a non USE_EXPAND too:
```
net-proxy/squid -* abi_x86_64 amd64 ecap elibc_glibc esi gnutls htcp kerberos kernel_linux ldap logrotate mysql pam perl postgres samba sasl snmp sqlite ssl ssl-crtd systemd userland_GNU wccp wccpv2
```

On running `smart-live-rebuild`:
```
[...]
WARNING:pkgcore:'/etc/portage/package.use/test', line 57: parsing error: token -* is not a valid use flag
[...]
```

cc @ferringb

arthurzam:
I think I managed to solve this with https://github.com/pkgcore/pkgcore/pull/397 and by reading [PMS](https://projects.gentoo.org/pms/8/pms.html#x1-560005.3.1).

PR #397, opened 2023-01-20
domain: support `-*` in use as previous removal

arthurzam:
Add support for `-*` and `USE_EXPAND: -*` syntax. Allow this syntax, and update incremental computation of resulting use combination.

Resolves: https://github.com/pkgcore/pkgcore/issues/393

### Notes

I'm not sure I've solved it in the ideal way, on the fast way. From multiple tests, it seems to work, but this part of pkgcore is still hard for me to understand (especially the change in `incremental_chunked`).

### DIFF
--- a/src/pkgcore/ebuild/misc.py
+++ b/src/pkgcore/ebuild/misc.py
@@ -66,8 +66,14 @@ def optimize_incrementals(sequence):
                 finalized.add(item)
 
 
-def incremental_chunked(orig, iterables):
+def incremental_chunked(orig: set[str], iterables):
     for cinst in iterables:
+        if "*" in cinst.neg:  # remove all previous set flags
+            orig.clear()
+        for flag in cinst.neg:
+            if flag.endswith("_*"):  # remove previous USE_EXPAND
+                drop = [f for f in orig if f.startswith(flag[:-2])]
+                orig.difference_update(drop)
         orig.difference_update(cinst.neg)
         orig.update(cinst.pos)
 
--- a/src/pkgcore/ebuild/domain.py
+++ b/src/pkgcore/ebuild/domain.py
@@ -69,11 +69,11 @@ def package_masks(iterable):
 
 def restriction_payload_splitter(iterable, post_process=lambda x: x):
     for line, lineno, path in iterable:
-        v = line.split()
+        pkg, *flags = line.split()
         try:
             # TODO: expand this invocation to allow threading token level validation down.
             # things like "is this a valid use flag?"
-            yield parse_match(v[0]), tuple(post_process(v[1:])), line, lineno, path
+            yield parse_match(pkg), tuple(post_process(flags)), line, lineno, path
         except ParseError as e:
             logger.warning(f"{path!r}, line {lineno}: parsing error: {e}")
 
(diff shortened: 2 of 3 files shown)


ferringb:
@arthurzam I'll take a look at this before you're up, pardon the lag on my end.

I'll also look at expanding the domain tests, although for some of it, I don't have any easy way to assert validation with out a bit of refactoring.

src/pkgcore/ebuild/misc.py
```diff
@@ -66,8 +66,14 @@ def optimize_incrementals(sequence):
                 finalized.add(item)
 
 
-def incremental_chunked(orig, iterables):
+def incremental_chunked(orig: set[str], iterables):
     for cinst in iterables:
+        if "*" in cinst.neg:  # remove all previous set flags
+            orig.clear()
+        for flag in cinst.neg:
+            if flag.endswith("_*"):  # remove previous USE_EXPAND
```
ferringb, line 74: I'm surprised this is safe; ChunkedDataDict wasn't written as a use specific thing.  In tracing the code, it doesn't look like anything other than 'USE' pathways use the chunkdatadict however.

ferringb:
@arthurzam looking through this, my rough thoughts are "merge it" but also a "let's rewrite this also".

Roughly:
* the tokenization logic just got a heck of a lot more complicated in regards to state management.  I'll see if I can suggest a refactoring rather than just a criticism.
* We're now threading something of a magic constant around in the form of `{use_expand}_*`; that sort of thing leads to no one understanding it in a year+.  I say that as a prime offender of that, just to be clear. ;)
* I'm not a huge fan of ChunkedDataDict becoming specialized to use negations, although that may not matter ultimately; in tracing modern portage conf, I'm not finding the old behaviour of things like `-*` doing incremental clearing of previous settings; things like licenses, keywords, etc, is where I recall that still being in used.
* We've got still got a negation bug for package.use that has `x y -* z`.  See https://github.com/ferringb/pkgcore/commit/1ebe88e2581ff85ad91df7f25410c829de7d8e09 for a test that captures this.  As best I can tell, that bug (and the equivalent use expand issue) has existed for a long while.

Either way, that's my thoughts; I don't much like this change but I'm hard pressed to provide something better at this juncture. ;)

arthurzam:
OK, after 2 weeks of me hoping a better solution without major rewrite of that area, I gave up. I will merge it as is, in hope of better solution in the future.

ferringb:
> OK, after 2 weeks of me hoping a better solution without major rewrite of that area, I gave up. I will merge it as is, in hope of better solution in the future.

Agreed, but pardon if you felt the need to wait for me (or a suggestion from me).  My phrasing was intended to convey "merge it, even if there are issues with it".

ferringb:
More to the point, in the future feel free to go forward; if I see an issue that warrants not doing something, I'll be clear (usually w/ the words "this breaks shit").  Anything less than that is just review discussion not intended to impede.

arthurzam:
> Agreed, but pardon if you felt the need to wait for me (or a suggestion from me). My phrasing was intended to convey "merge it, even if there are issues with it".

Oh, no problems. No offense, but as you said to me when you introduced yourself, you are a "random" dev, so never expect anything (but I'm happy for anything you bring).

> More to the point, in the future feel free to go forward; if I see an issue that warrants not doing something, I'll be clear (usually w/ the words "this breaks shit"). Anything less than that is just review discussion not intended to impede.

Yeah, I wasn't waiting for you, I was more trying myself, and after 2 weeks I gave up myself to find a better solution for now, and I merged. I'm also not the fastest developer, and try to work on the whole stack. For me pkgcheck feels the most rewarding, pkgdev is the hardest (cause new tools always require huge graph work) and pkgcore is the weirdest ;-) So sometimes things "seem" stuck until suddenly I do stuff.

thesamesam:
Yeah, I just took it as "eh, none of us had a better idea, let's go ahead", nothing bad. I was hoping to look at this and come up with something better and couldn't 😢

PR #398, opened 2023-02-05
config cleanup: removal of unused/deprecated functionality

ferringb:
The commit's are all detailed in full for what they're doing, but this PR does roughly thus:
* Remove 'incremental' logic from configuration objects.  Back in the day, the intent was to have `-x86 x86 x64 -x64` render as `x86` as part of instantiation; portage configuration got convoluted enough to render this idea massively dead in the water.  Thus remove it, since all incremental logic is handled at the domain or config parsing level.
* remove prepend/append manipulation logic from configs object instantiation.  These exist purely to support incrementalism.  They still work, but they add significant complexity, thus remove 'em.
* cleanup migration/deprecation for how config objects are accessed.  Back in the 2.3/2.4 era, the central object sections were access directly on the central manager; this lead to conflicts between the object's protocol/members, and sections being instantainted.  Said another way, all central 'instantiations' were moved into an attribute namespace that couldn't conflict in a reasonably fashion.  TL:DR: access objects via central.objects.<type>.<name> rater than central.<type>.<name> .  Additionally convert the old form of access to a warning, since python makes this easy now.
* finally, some minor refactoring cleanup.  Things like removing duplicated methods/logic folks added, or reducing configuration instantiation hints to the actual class that does it.


src/pkgcore/config/basics.py
```diff
@@ -597,9 +440,9 @@ def convert_hybrid(central, value, arg_type: str):
 
 # "Invalid name" (pylint thinks these are module-level constants)
 # pylint: disable-msg=C0103
-HardCodedConfigSection = partial(FakeIncrementalDictConfigSection, convert_asis)
-ConfigSectionFromStringDict = partial(FakeIncrementalDictConfigSection, convert_string)
-AutoConfigSection = partial(FakeIncrementalDictConfigSection, convert_hybrid)
+HardCodedConfigSection = partial(DictConfigSection, convert_asis)
```
ferringb, line 443: later PR's should cleanup the type conversion logic.  If you trace through those, they're fairly fucking gross, but they were necessary shims (at the time) to make things like `pquery --add-config` work properly w/ the frameworks in use at the time.

They can be improved, is the TL;DR.

arthurzam (approved):
Looks great, thank you for your work and great explanations in each commit.

PR #402, opened 2023-03-20
CI: add Python 3.12

thesamesam:


arthurzam:
`lxml` dep is cause us to fail on py3.12

arthurzam:
Can you please rebase, I'm suspecting it would pass now

thesamesam:
Done. Let's hope so. Annoying that we can test it fine in Gentoo but we're stuck with CI here :(

arthurzam (approved):

PR #406, opened 2023-06-29
fix missing errors when all ebuilds for package are masked bad

echuber2:
Fixes https://github.com/pkgcore/pkgdev/issues/55

If all of a package's ebuild files have been masked bad, that package gets filtered out of the iterator for the previous loop in this function, so a followup loop over the masked bad packages can help to emit the unreported errors. The commit adds an extra warning message about the edge case, which may not be necessary.

Co-authored-by: @zhuyifei1999


thesamesam:
Many thanks for this. It's very much appreciated, this has been a really annoying bug for some time.

Issue #411, opened 2023-08-28
Implement stabilization groups parsing


arthurzam:
1. everything is located under "metadata/stabilization-groups/"
2. We search recursively as much as needed, so all files are included in any depth.
3. Group name plays similarly to path, so here the group name would be "@gnome/vte" (at least for `pkgdev bugs` invocations). By using full path, we are safe from name collisions.
4. The file itself uses similar format to our various profiles files. Ignore white-spaces before and after, "#" as a comment. Only one "cat/pkg" per line.
5. I think for now we can go without mandatory copyright header...

PR #412, opened 2023-08-28
ebuild.repository: add support for stabilization groups

arthurzam:
This is a special metadata files that can be used to group packages for stabilization bugs.

Resolves: https://github.com/pkgcore/pkgcore/issues/411

### DIFF
--- a/src/pkgcore/ebuild/repository.py
+++ b/src/pkgcore/ebuild/repository.py
@@ -16,6 +16,7 @@ from snakeoil import chksum, klass
 from snakeoil.bash import read_dict
 from snakeoil.containers import InvertedContains
 from snakeoil.data_source import local_source
+from snakeoil.fileutils import readlines_utf8
 from snakeoil.mappings import ImmutableDict
 from snakeoil.obj import make_kls
 from snakeoil.osutils import listdir_dirs, listdir_files, pjoin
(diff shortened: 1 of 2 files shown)


src/pkgcore/ebuild/repository.py
```diff
@@ -670,6 +673,27 @@ def deprecated(self):
...
         return packages.OrRestriction(*pkg_deprecated)
 
+    @klass.jit_attr
+    def stabilization_groups(self):
+        """Return a mapping of stabilization groups to packages."""
+        stabilization_groups = {}
+        base_dir = pjoin(self.location, "metadata", "stabilization-groups")
+        for dirname, _dirs, files in os.walk(base_dir):
+            dirbase = dirname.removeprefix(base_dir)
+            for file in files:
```
mgorny, line 683: I'd rather you require a specific suffix. Otherwise we're going to end up with the mess of reading backup files accidentally.

arthurzam: We don't have any suffix defined for those, but I do plan to verify this in pkgcheck with error level (so it breaks the QA bot). I prefer those names be as simple as possible.

mgorny: Just make them `*.group`, that should be simple and clean, and it will avoid issues in the future. We've made a lot of recursive directories in Portage with no suffixes, and you know how complex handling these has become.

PR #413, opened 2023-08-29
new release 0.12.23

arthurzam:


thesamesam (approved):
Thanks!

Maybe for completeness we could link to the ML post about stable groups.

thesamesam:
I meant in terms of "this is the spec for it", not "you can go use it now"

PR #417, opened 2023-12-02
Remove multi-level categories, misc minor cleanup

ferringb:
High level: pkgcore was originally built to allow multi-level categories.  Think media/sound/alsactl , and you're in range.

Being this will never happen in gentoo, might as well rip it out to simplify.

The other changes are similar simplification- either replacing adhoc mocks w/ existing solutions, or cleanup of signatures as I've been re-auditing the code.

See the raw commits for further ellucidation.


src/pkgcore/repository/multiplex.py
```diff
@@ -91,25 +91,15 @@ def __init__(self, *trees, repos=()):
                 )
         self.trees = trees
 
-    def _get_categories(self, *optional_category):
+    def _get_categories(self):
         d = set()
         failures = 0
```
ferringb, line 96: this can be simplified into a bool

arthurzam (approved):

ferringb:
@arthurzam how are you merging my PR's btw?  I'm asking because it's been fairly hit or miss for me doing rebases, and I'm not quite understanding why.  I don't see content changing- sha1 does- which I assume is from you cherry-picking.

Either way, I'd appreciate a quick dump of how you merge these changes into gentoo's master.  I tend to generate a shiteload of patches sporadically and juggling them post merge is something of a slowdown/PITA... which may just be workflow issues on my end.

arthurzam:
So I'm using `pram` utility to add my Signed-off to your commits, and also sign with my GPG (to comply with Gentoo's GLEPs and such, similar flow to merging to ::gentoo tree)

What you want to do is to use to rebase is `git rebase --onto=upstream_pkgcore/master ${SHA1_COMMIT}`, where the `SHA1_COMMIT` is the commit id of your sha1 (for example in this PR it is 32a67e1). I recommend reading some more about `git rebase --onto` - this is very nice tool.

ferringb:
Regarding how to use `git rebase`, I'm familiar; if you don't know of `git rebase -i`, well, it's been my crutch for the last decade to make it look like all my commits are pristine.

As to the issue, I ran it down and it comes down to my configs being fairly fucking old- I had an old merge strategy enabled which wasn't properly handling the commit message changing, but the delta being the same.

Thanks for confirming your workflow- it verified it had to be on my end.

Issue #419, opened 2023-12-24
patom: no version when no =


mgorny:
```
$ patom -F '%{fullver} %{version}' =foo/bar-11-r3
11-r3 11
$ patom -F '%{fullver} %{version}' foo/bar-11-r3
<unset> <unset>
```

It seems a bit weird that it accepts versioned atom without `=` but ignores the version.

ferringb:
```
ferringb@frith ~/gentoo/pkgcore $ python -c 'from pkgcore.ebuild.atom import atom;print(atom("foo/bar-11-r3").package)'
bar-11-r3
```
I suspect either this is a gap in PMS atom syntax, or there is a gap in pkgcore's atom parsing.

ferringb:
There's a parsing bug in this- which I have a PR for- but this '=' parsing is intentional from @arthurzam (see https://github.com/pkgcore/pkgcore/commit/d5e0e76bf311d5c2718d6db8d97c7e7e8dd589b5#diff-61a8838eb341cc91a984511943886afe9ac969b81c65faa292fd39adfc39931eR11 ).  Personally I think it should be removed- @arthurzam your thoughts?

arthurzam:
> There's a parsing bug in this- which I have a PR for- but this '=' parsing is intentional from @arthurzam (see [d5e0e76#diff-61a8838eb341cc91a984511943886afe9ac969b81c65faa292fd39adfc39931eR11](https://github.com/pkgcore/pkgcore/commit/d5e0e76bf311d5c2718d6db8d97c7e7e8dd589b5#diff-61a8838eb341cc91a984511943886afe9ac969b81c65faa292fd39adfc39931eR11) ). Personally I think it should be removed- @arthurzam your thoughts?

This was intentional, and only used in the `patom` tool, since it should be simpler for consumers, even if not ideally correct.

arthurzam:
With @ferringb fix merged, I usecase passes now, so I consider the issue closed.
If someone disagrees, feel free to reopen :)

PR #420, opened 2023-12-25
tighten CPV parsing to disallow revisions as part of the package name

ferringb:
This adds tests for CPV to catch `foo/bar-11-r3` treating `bar-11-r3` as a package name due to it not checking for trailing revisions, just versions.  This also adds a redundant assertion in atom parsing to ensure CPV changes don't regress the atom parsing.

This just fixes a tangential parsing issue spotted while looking at #419 ; it doesn't resolve the design choice for that.


arthurzam (approved):
Thank you, I'll merge it soon.

Issue #421, opened 2023-12-27
pmaint regen: error: x11-drivers/xf86-video-r128-6.12.1: invalid package name       


thesamesam:
In `x11-drivers/xf86-video-r128` at ::gentoo commit `b923be7a4099bd8a50ca0991fe3c88f7966eab7d`, I get:
```
$ pkgcheck scan
pkgcheck scan: error: Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/pkgcore/ebuild/atom.py", line 298, in __init__
    sf(self, "_cpv", cpv.CPV(self.cpvstr, versioned=bool(self.op)))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/pkgcore/ebuild/cpv.py", line 333, in __init__
    raise InvalidCPV(cpvstr, "invalid package name")
pkgcore.ebuild.errors.InvalidCPV: x11-drivers/xf86-video-r128-6.12.1: invalid package name

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/pkgcheck/pipeline.py", line 231, in _run
    self._queue_work(sync_pipes, work_q)
  File "/usr/lib/python3.12/site-packages/pkgcheck/pipeline.py", line 173, in _queue_work
    for restrict in versioned_source.itermatch(restriction):
  File "/usr/lib/python3.12/site-packages/pkgcheck/sources.py", line 398, in itermatch
    new = self.keyfunc(pkg)
          ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/pkgcheck/packages.py", line 34, in versioned_atom
    return atom.atom(f"={self}")
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/snakeoil/caching.py", line 114, in __call__
    instance = super(WeakInstMeta, cls).__call__(*a, **kw)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/pkgcore/ebuild/atom.py", line 300, in __init__
    raise errors.MalformedAtom(orig_atom) from e
pkgcore.ebuild.errors.MalformedAtom: invalid package atom: '=x11-drivers/xf86-video-r128-6.12.1'
```

[...]

PR #422, opened 2023-12-27
fix: tweak CPV parsing rules to match PMS.

ferringb:
Bug #421 captures this; pkgcore was allowing version components as the package name if it was suffixed by a revision; the fix for that incorrectly limited a package name that has a trailing revision, but *no version syntax* in the name.

This commit just refactors the revision check to also force a check for a leading version if revision-like is found.

### DIFF
--- a/src/pkgcore/ebuild/cpv.py
+++ b/src/pkgcore/ebuild/cpv.py
@@ -42,7 +42,12 @@ def isvalid_pkg_name(chunks):
     # chunk, i.e. at least one hyphen
     if len(chunks) == 1:
         return True
-    return not (isvalid_version_re.match(chunks[-1]) or isvalid_rev(chunks[-1]))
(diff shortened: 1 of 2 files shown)


src/pkgcore/ebuild/cpv.py
```diff
@@ -42,7 +42,12 @@ def isvalid_pkg_name(chunks):
     # chunk, i.e. at least one hyphen
     if len(chunks) == 1:
         return True
-    return not (isvalid_version_re.match(chunks[-1]) or isvalid_rev(chunks[-1]))
+    if isvalid_version_re.match(chunks[-1]):
```
ferringb, line 45: I kind of hate the logic flow here, even if it captures it correctly.  Suggestions welcome, just keep in mind this is something of a 'fast path' for package name validation, so it's best to align the checks accordingly.

arthurzam (approved):
thank you

PR #424, opened 2024-01-12
fix: child repository not inheriting license groups when it defines none

ferringb:
If the child repository defines no license groups, the previous code would exit out before it integrated the parents license groups.

This fixes that, and adds tests for it.


ferringb:
@mgorny this shouldn't affect any pkgcheck infra- in fact it should fix anything that was flagging this- but I'd appreciate an ack on your end since this is a mildly subtle fix.

ferringb:
pushing a tweak to make codecov happy.

arthurzam (approved):

PR #428, opened 2024-01-23
Hygenic test fix + repository typing cleanup + fixes

ferringb:
The commits carry detailed info for each change.  Summarizing:
* Downstream usage of the pytest plugin had assumption about environment hygiene ($HOME being sanitized) that wasn't true.  That's fixed.  That primarily impacted test validation of pkgdev.
* Repository prototype as converted to an ABC so any derivatives that don't override abstract methods get flagged; both filtered (actually broke) and configured (false positive, but best to be explicit) were caught via this.
* * For filtered, the `.{category,package,versions}` API was broken via `super().__init__()` not being invoked.  This is fixed, and proper implementations of the underlying `_get_.*` methods was added, along with tests.
*  `prototype.tree._get_{category,packages,versions}` had a requirement that child implementations return the results in an immutable form.  This is begging for a bug; instead, accept either an immutable or an iterable, and within the ABC *enforce* it as immutable.  For implementations that already returned immutable, this is effectively a noop- `tuple(some_tuple)` returns `some_tuple` w/out creating a new one, as does frozenset.

### DIFF
--- a/src/pkgcore/pytest/plugin.py
+++ b/src/pkgcore/pytest/plugin.py
@@ -41,6 +41,8 @@ class GitRepo:
                 self.add(pjoin(self.path, ".init"), create=True)
 
     def run(self, cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, **kwargs):
+        env = os.environ.copy()
+        env["HOME"] = self.path
         return subprocess.run(
             cmd,
             cwd=self.path,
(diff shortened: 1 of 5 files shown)


ferringb:
I pulled the offending commits that impacted pkgdev; I'm unclear why that's not showing in my local testing but I'll run it down (I expect pebkac rather than test hygiene).  Either way, they'll come in a later PR.

arthurzam (approved):
Thank you very much.

PR #430, opened 2024-01-25
fix: tox.ini since it's not been used in ages.

ferringb:
This broke whenever the repo went to pyproject.toml .

On the one hand, I like enumerating what devs have to invoke locally rather than relying on GH.

On the other hand, clearly no one is using this.

This fixes it for current usage.  If it's to continue, then GH and tox envlist's should be fused.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,25 +1,22 @@
 [tox]
-envlist = py38, py39, py310, py311, py312
+envlist = py310, py311, py312
 [testenv]
 # force latest virtualenv/pip
 download = true
-deps = -rrequirements/tox.txt
+extras = test
 commands =
 	pytest --cov {posargs:-v}
 
 # build docs
 [testenv:docs]
 skip_install = true
-deps =
-	-rrequirements/dev.txt
-	-rrequirements/docs.txt
+extras = doc
 commands =
 	python setup.py build_docs
 
 # build dist files
 [testenv:dist]
 skip_install = true
-deps = -rrequirements/dist.txt
 commands =
 	python setup.py sdist
 	python setup.py bdist_wheel


ferringb:
@thesamesam , @mgorny , @arthurzam : how do y'all do your testing prior to pushing and having an "oh shit" moment?

I'm asking since tox was clearly dead; it's fixed, and I'm inclined to wire GH workflow to this, but if no one values tox... well.  You get the idea.

Preferences?

arthurzam:
> @thesamesam , @mgorny , @arthurzam : how do y'all do your testing prior to pushing and having an "oh shit" moment?

I have a shared workspace where all the stack packages are installed as editable in a virtual env. When it comes to test, I call `pytest -n 4`. If it failed, I call `pytest` and then see the failures. For cross python version support, I just remember what was included when.

> I'm asking since tox was clearly dead; it's fixed, and I'm inclined to wire GH workflow to this, but if no one values tox... well. You get the idea.
> 
> Preferences?

I myself don't value so much tox, since it is complicated and weird for me. But I think @mgorny is using tox, so I don't know... Maybe a small pipeline to just verify it works is nice (this is GitHub's CI money, so adding more doesn't hit us at all).

ferringb:
@arthurzam in terms of your pipeline suggestion, the GH matrix's actually can consume the output of a previous build step.  My thought was to add a step that extracts the env list out of tox.ini and then uses that for the python version matrix for GH.

That adds complexity that I feel is only warranted if tox has value.  I *do* like the ability to run things in parallel in separate venvs for hygenic reasons, but I kind of hate interacting w/ tox for reasons you referred to.  Is there any alternative to it?

thesamesam:
I do roughly what Arthur described but I do use tox when I'm done iterating as well. I think @mgorny is the biggest user of tox. I think the reason I didn't notice until now is I haven't done much pkg* work recently ;)

PR #431, opened 2024-01-25
Correct fix intersphinx doc linkage, automate maintenance of it.

ferringb:
intersphinx is used to do linkage between projects- pkgcore referring to snakeoil documentation for example.

These links are long since old/dead, thus need fixing.

See https://github.com/pkgcore/pkgcore/actions/runs/7656999844/job/20874145430 for an example of the failure.


ferringb:
Same comment as in snakeoil, use https://pypi.org/project/seed-intersphinx-mapping/ to automate this.

arthurzam (approved):

PR #435, opened 2024-06-11
ebuild.profiles: Fix case where a parent path omits the repo identifier

chewi:
It is valid to have a parent path like `:path/to/profile` where the repo identifier is missing. This refers to a path in the current repo.


thesamesam:
To be clear, it's valid (presumably) for `portage-2`. Maybe link if you can to where Portage declares this is supposed to be supported?

chewi:
It states this in `man portage`.

> If `layout.conf` is  new  enough, you can also use the `<repo>:<path>` syntax.  The `<repo>` is the same string as is stored in the `repo_name` file **(or omitted to refer to the current repo)**, and `<path>` is a subdir starting at `profiles/`.

I've now mentioned it in the commit message.

ferringb:
LGTM

chewi:
Fixed the Signed-off-by address.

PR #446, opened 2025-05-31
tests: Wrap partial() in staticmethod(), in attributes for py3.14

mgorny:
Wrap `partial()` uses in class attributes in `staticmethod()`, as required for them to work correctly in Python 3.14.

> functools.partial is now a method descriptor. Wrap it in staticmethod()
> if you want to preserve the old behavior.

(https://docs.python.org/3.14/whatsnew/3.14.html#changes-in-the-python-api)

### DIFF
--- a/tests/ebuild/test_profiles.py
+++ b/tests/ebuild/test_profiles.py
@@ -787,7 +787,7 @@ class TestPortage1ProfileNode(TestPmsProfileNode):
         ]
     )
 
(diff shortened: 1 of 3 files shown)


thesamesam (approved):
It looks OK to me but I'm not confident with this at all, so up to you if you want to wait for @arthurzam or @ferringb.

PR #456, opened 2025-11-22
EAPI 9: Update bash_compat from 5.2 to 5.3

ulm:
Bug: https://bugs.gentoo.org/946193


ulm:
Some checks fail, but looks like this is a problem with the bash version of the runner (ancient Ubuntu) being outdated, not a problem with the code.

PR #464, opened 2026-02-12
fetch.custom: Allow for partial resuming

swagtoy:
This will attempt to download all files as XXX._partial_. The file
will rename to XXX upon success.

Correlatively, this also fixes the issue where interrupted
downloads (i.e. pkgdev manifest) ended up counting as fetched,
requiring the user to manually delete them to continue.

### DIFF
--- a/src/pkgcore/fetch/custom.py
+++ b/src/pkgcore/fetch/custom.py
@@ -105,6 +105,7 @@ class fetcher(base.fetcher):
         if not isinstance(target, fetchable):
             raise TypeError(f"target must be fetchable instance/derivative: {target}")
 
+        partial_fetch = False
         path = pjoin(self.distdir, target.filename)
         uris = iter(target.uri)
         last_exc = RuntimeError("fetching failed for an unknown reason")
@@ -112,10 +113,18 @@ class fetcher(base.fetcher):
         if self.userpriv and is_userpriv_capable():
             spawn_opts.update({"uid": portage_uid, "gid": portage_gid})
 
+        if os.path.exists(f"{path}._partial_"):
+            command = self.resume_command
+            partial_fetch = True
+
         for _attempt in range(self.attempts):
             try:
-                self._verify(path, target)
-                return path
+                if partial_fetch:
+                    # Can't verify here, but we still need to keep fetching
+                    partial_fetch = False
+                else:
+                    self._verify(path, target)
+                    return path
             except errors.MissingDistfile as exc:
                 command = self.command
                 last_exc = exc
@@ -135,9 +144,15 @@ class fetcher(base.fetcher):
             # the loop handles this. In other words, don't trust the external
             # fetcher's exit code, trust our chksums instead.
             try:
-                spawn_bash(
-                    command % {"URI": next(uris), "FILE": target.filename}, **spawn_opts
-                )
+                if 0 == spawn_bash(
+                    command % {"URI": next(uris), "FILE": f"{target.filename}._partial_"}, **spawn_opts
+                ):
+                    os.rename(f"{path}._partial_", path)
+            except FileNotFoundError:
+                # If a _partial_ rename fails then that should imply it's
+                # completed.
+                assert os.path.exists(path)
+                pass
             except StopIteration:
                 raise errors.FetchFailed(
                     target.filename, "ran out of urls to fetch from"


eli-schwartz (changes requested):

swagtoy:
gonna change this to just start downloading straight to ._partial_ entirely, actually.

src/pkgcore/fetch/custom.py
```diff
@@ -105,17 +105,26 @@ def fetch(self, target: fetchable):
...
 
+        partial_fetch = False
         path = pjoin(self.distdir, target.filename)
         uris = iter(target.uri)
         last_exc = RuntimeError("fetching failed for an unknown reason")
         spawn_opts = {"umask": 0o002, "env": self.extra_env}
         if self.userpriv and is_userpriv_capable():
             spawn_opts.update({"uid": portage_uid, "gid": portage_gid})
 
+        if os.path.exists(f"{path}._partial_"):
```
ferringb, line 116: What is portage's standard naming for partial files?

src/pkgcore/fetch/custom.py
```diff
@@ -105,17 +105,26 @@ def fetch(self, target: fetchable):
...
+            command = self.resume_command
+            partial_fetch = True
+
         for _attempt in range(self.attempts):
             try:
-                self._verify(path, target)
-                return path
+                if partial_fetch:
+                    # Can't verify here, but we still need to keep fetching
+                    partial_fetch = False
```
ferringb, line 124: Unless I'm missing something, `partial_fetch` boolean does nothing- if it's enabled, it's immediately disabled.  Right?

src/pkgcore/fetch/custom.py
```diff
@@ -135,9 +144,15 @@ def fetch(self, target: fetchable):
             # the loop handles this. In other words, don't trust the external
             # fetcher's exit code, trust our chksums instead.
             try:
-                spawn_bash(
-                    command % {"URI": next(uris), "FILE": target.filename}, **spawn_opts
-                )
+                if 0 == spawn_bash(
+                    command % {"URI": next(uris), "FILE": f"{target.filename}._partial_"}, **spawn_opts
+                ):
+                    os.rename(f"{path}._partial_", path)
```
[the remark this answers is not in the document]
ferringb: There's not much one can do here, imo.  This will be some form of EnvironmentError- IOError, OSError, etc.  I'd expect permissions to be the actual failure, if any.

What scenarios are you thinking about?

src/pkgcore/fetch/custom.py
ferringb: Also, what happens if a partial and the actual target file exists?  I'd assume a verify of the 'target'- use it and wipe partial if there is no lock- is correct.

If it's not valid, wipe the target, resume the partial, imo.

ferringb (commented):
This is a first pass- I've not pulled the code in IDE and done tracing, and I grok this is something of a draft looking for feedback.  Pardon the lag, in light of that.

A few thoughts:

If this is resuming a partial, there is a scenario where the hash has changed for the file- validly, including the scenario of the size changing.  It's corner case as hell, but I know it's happens time to time from upstreams or failed release processes.

To wit, if the resume command 'finishes' for a partial, and it doesn't pass checksums, is the correct behavior to wipe and fetch from scratch?

Also: how does this interact with any FS level locking the fetcher has in place, if any?  flock, stuff like that.

As mentioned, aligning w/ what portage does for the file name would be desirable.

eli-schwartz:
> If this is resuming a partial, there is a scenario where the hash has changed for the file- validly, including the scenario of the size changing. It's corner case as hell, but I know it's happens time to time from upstreams or failed release processes.
> 
> To wit, if the resume command 'finishes' for a partial, and it doesn't pass checksums, is the correct behavior to wipe and fetch from scratch?

I don't see why not. Trying to preserve partial chunks from the current session and backfill the original partial's ranges sounds very painful.

I don't think this should really happen anyway. If upstream has re-rolled the release tarball we should use arrow renames to give the new tarball a unique name. (Among other things this guarantees that Manifest will actually get regenerated rather than downloading the old tarball from GENTOO_MIRRORS by accident.) Failure to do that is arguably an ebuild author bug. And even then the result is still correct as far as the distfile fetching is concerned -- whatever is in Manifest will get downloaded from scratch until things match, and there is merely wasted bandwidth as a side effect (not permanently destructive).

src/pkgcore/fetch/custom.py
swagtoy: Yes; it'll basically skip over the initial verification step.

src/pkgcore/fetch/custom.py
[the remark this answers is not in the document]
swagtoy: I'm talking about the resume command here failing, not the rename part, I just don't know what I should do if an error occurs for that.

PR #465, opened 2026-03-21
pkgcore: profiles: use ProfileNode for base profile

CyberTailor:
I don't understand why is `RepoConfig.base_profile` instantiated as [`EmptyRootNode`](https://pkgcore.github.io/pkgcore/api/pkgcore.ebuild.profiles.html#pkgcore.ebuild.profiles.EmptyRootNode). Why couldn't it be [`ProfileNode`](https://pkgcore.github.io/pkgcore/api/pkgcore.ebuild.profiles.html#pkgcore.ebuild.profiles.ProfileNode) instead?

Using `EmptyRootNode` leads to unexpected (to me) behavior. For example, ::guru has `profiles/package.use.mask` but `UnconfiguredTree.config.base_profile.pkg_use_mask` is empty.

### DIFF
--- a/src/pkgcore/ebuild/profiles.py
+++ b/src/pkgcore/ebuild/profiles.py
@@ -910,7 +910,7 @@ class OnDiskProfile(ProfileStack):
     def stack(self):
         l = ProfileStack.stack.function(self)
         if self.load_profile_base:
-            l = (EmptyRootNode._autodetect_and_create(self.basepath),) + l
+            l = (ProfileNode._autodetect_and_create(self.basepath),) + l
         return l
 
     @klass.jit_attr
--- a/src/pkgcore/ebuild/profiles.py
+++ b/src/pkgcore/ebuild/profiles.py
@@ -1,7 +1,6 @@
 __all__ = (
     "ProfileError",
     "ProfileNode",
-    "EmptyRootNode",
     "OnDiskProfile",
     "UserProfile",
 )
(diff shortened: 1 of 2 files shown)


ferringb:
I introduced EmptyRootNode in e89532aee02090bb4e95b1cb7ca9594def2e7abe .

I've got theories, but I'll buy a beer for whoever conclusively figures out *why* I did that.  My assumption is that it was to allow isinstance for profile linked list walk termination- back then multi-inheritance of profiles wasn't allowed (IIRC).

@CyberTailor - can you wipe the dead class and whatever tests related?  I left a comment near code you're touching, which I'd appreciate some tracing on.  If that tracing turns up a rats nest, lets move it over to a ticket.  There's a code-smell in that area that I suspect would allow simplification, or is hiding a corner case no one has flushed yet.

Deleting the dead code is the only part missing for a merge, from my standpoint.

ferringb:
See https://github.com/pkgcore/pkgcheck/blob/5986de6fb0df0531109105a9359a546f37a9f38c/src/pkgcheck/addons/__init__.py#L173-L192 ; that's turning on the `use EmptyRootNode` which suppresses the profile root data (the situation you mentioned).   That code then does it's own poking at the repo config.

also:

https://github.com/pkgcore/pkgcheck/blob/5986de6fb0df0531109105a9359a546f37a9f38c/src/pkgcheck/addons/profiles.py#L187 that also enables this option.

The history on the `load_profile_base` change maps back to e89532aee02090bb4e95b1cb7ca9594def2e7abe , and that time frame was dealing w/ far different profile rules.  It's very possible what this 'did', no longer matters, however the pkgcheck UseAddon and ProfileAddon are seriously twisty and I'm very much not clear on the implications of your proposed change.

I'm not really comfortable merging it without that understanding; to be clear, that's not a good thing.  It should be clear, but the pkgcheck profile caching code was always *seriously* complex.

What I can state in pulling old history, is the introduction of `load_profile_base` proceeded the first commit of pkgcheck by ~5m- however it wouldn't shock me if I made the pkgcore side change in support of the pkgcheck code I was developing on the side.

What's your thoughts on a path forward here?  We could also kick this over to a ticket for future spelunking and cleanup.

